PR #7, opened 2016-03-04
Read params from ini file

Axik:
@datarobot/predictions

### DIFF
--- a/datarobot_batch_scoring/batch_scoring.py
+++ b/datarobot_batch_scoring/batch_scoring.py
@@ -891,7 +891,7 @@ def main(argv=sys.argv[1:]):
     if conf_file:
         file_args = parse_config_file(conf_file)
         parsed_args.update(file_args)
-    pre_parsed_args = {k: v for k, v in parser.parse_args()._get_kwargs() if v is not None}
+    pre_parsed_args = {k: v for k, v in vars(parser.parse_args(argv)).items() if v is not None}
     parsed_args.update(pre_parsed_args)
     level = logging.DEBUG if parsed_args['verbose'] else logging.INFO
     configure_logging(level)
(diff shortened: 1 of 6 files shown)


datarobot_batch_scoring/batch_scoring.py
```diff
@@ -889,47 +886,51 @@ def main(argv=sys.argv[1:]):
     parser.add_argument('-n', '--no', dest='prompt', action='store_false',
                         help="Always answer 'no' for user prompts")
 
-    parsed_args = parser.parse_args()
-    level = logging.DEBUG if parsed_args.verbose else logging.INFO
+    parsed_args = {}
+    conf_file = get_config_file()
+    if conf_file:
+        file_args = parse_config_file(conf_file)
+    parsed_args.update(file_args)
```
osnatil: is file_args defined in case conf_file is False?

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -889,47 +886,51 @@ def main(argv=sys.argv[1:]):
...
+    pre_parsed_args = {k: v for k, v in parser.parse_args()._get_kwargs() if v is not None}
```
osnatil: Are we interested in all the keys that their value is not None?

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -889,47 +886,51 @@ def main(argv=sys.argv[1:]):
...
+
+    concurrent = int(parsed_args['n_concurrent'])
+    dataset = parsed_args['dataset']
+    n_samples = int(parsed_args['n_samples'])
+    delimiter = parsed_args.get('delimiter')
+    resume = parsed_args['resume']
+    out_file = parsed_args['out']
+    datarobot_key = parsed_args.get('datarobot_key')
+    pwd = parsed_args['password']
+    timeout = int(parsed_args['timeout'])
```
osnatil: wouldn't a trafaret be a better use here?

datarobot_batch_scoring/batch_scoring.py
Axik: good catch, yes

datarobot_batch_scoring/batch_scoring.py
Axik: basically, yes

datarobot_batch_scoring/batch_scoring.py
asvetlov: I believe just specifying `type` in `add_argument()` does all work.

But I ok with making the change in separate PR.

asvetlov:
LGTM

Axik:
@asvetlov

asvetlov:
LGTM, please merge.
I have some ideas on future improvements but the PR is big enough, better to fix the current state.

PR #24, opened 2016-05-19
sniff more text, give meaningful exception on failure

mckaymatt:
the standard lib CSV module kept rising `Error: Could not determine delimiter` because it didn't have enough data. I raised  the sample to 1024*1024 and that should be enough for some wide datasets. I tested it out and sniffing that much data takes about 70ms. I think that's okay for batch scoring.

The other problem was that providing a `--delimiter` didn't help when the sniffer failed, and that was very frustrating. Now, even if the Sniifer fails, the batch-scoring script at least tries to run using the delimiter provided by the user. If it fails and the didn't provide a delimiter, they are prompted to try with a delimiter.


Axik:
Did you take a look on travis failures? There is a couple of pep8 errors

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -80,7 +80,15 @@ def __iter__(self):
         # handle unix tabs
         with self.open() as csvfile:
             sniffer = csv.Sniffer()
-            dialect = sniffer.sniff(csvfile.read(1024).decode('latin-1'))
+            try:
+                dialect = sniffer.sniff(
+                    csvfile.read(1048576).decode('latin-1'))
```
pprett: @mckaymatt 1MB instead of 1KB?

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -80,7 +80,15 @@ def __iter__(self):
...
+            except Exception as e:
+                if e.message == 'Could not determine delimiter':
+                    if sep is None:
+                        raise
+                    pass
+                raise
```
pprett: that looks a bit convoluted.. what are we raising when?

mckaymatt:
I am going to close this PR for now because I have spent a bit more time on the problem and think I have a better solution. I need to test it out first though.

PR #25, opened 2016-05-20
Shovel process

pprett:
Throughput enhancements:
 - separate process for disk IO and request payload serialization
 - avoid codecs.getreader due to IO bottleneck
 - dont parse CSV (fail fatally on multiline csv)
 - `--multiline` mode (to be renamed)

Todo:
  - windows testing
  - `--keep_cols` resolution

@Axik pls review


datarobot_batch_scoring/batch_scoring.py
```diff
@@ -1,22 +1,26 @@
...
 import operator
 import os
 import shelve
 import sys
 import threading
 import hashlib
-from functools import partial, reduce
-from pprint import pformat
+from functools import partial
+from functools import reduce
```
alexeyraspopov: This seems to be not used anywhere.

datarobot_batch_scoring/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '1.6.0a4'
+__version__ = '1.6.1a2'
```
alexeyraspopov: Since new API (functionality?) is introduced, Minor version should be bumped, not patch (in case if we use [semver](http://semver.org/)).

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -41,53 +45,126 @@ class ShelveError(Exception):
...
+        self.fd = fd
+        self.sep = sep
+        self.dialect = dialect
+        self.reader = csv.reader(self.fd, dialect, delimiter=sep)
+        self.header = next(self.reader)
+        self.fieldnames = [c.strip() for c in self.header]
+
+    def __iter__(self):
+        self.fd.seek(0)
+        self.reader = csv.reader(self.fd, self.dialect, delimiter=self.sep)
```
alexeyraspopov: Should this be defined here since it's already defined in the `__init__`?

datarobot_batch_scoring/__init__.py
pprett: makes sense -- thx

datarobot_batch_scoring/batch_scoring.py
pprett: i think the reader doesnt support a reinitialize -- it will consume lines from the upstream stream (fd) but buffers them internally to support multiline. So if you seek the upstream fd we should reset the reader.

datarobot_batch_scoring/batch_scoring.py
pprett: thx

Makefile
```diff
@@ -15,7 +15,7 @@ test: .install flake8
 	py.test -v tests/
 
 cov cover coverage: .install flake8
-	py.test -v --cov=datarobot_batch_scoring \
+	py.test -v -x --cov=datarobot_batch_scoring \
```
Axik: it'll fail after first failed test. I don't think we want this at travis builds (annoying if you have two failed tests and it'll be shown in two builds, not one)

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -41,53 +45,126 @@ class ShelveError(Exception):
...
     Yields
     ------
     batch : Batch
-        The next batch
+        The next batch. A batch holds the data to be send already
+        in the form that can be passed to the HTTP request.
     """
 
-    def __init__(self, dataset, n_samples, n_retry, delimiter, ui):
+    def __init__(self, dataset, n_samples, n_retry, delimiter, ui, fast_mode):
```
Axik: should we support to pass encoding here too?

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -41,53 +45,126 @@ class ShelveError(Exception):
...
         self.dataset = dataset
         self.chunksize = n_samples
         self.rty_cnt = n_retry
         self.sep = delimiter
         self._ui = ui
+        self.fast_mode = fast_mode
+        self.encoding = 'utf-8'
```
Axik: if it's internal only variable, it should be `_encoding`

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -41,53 +45,126 @@ class ShelveError(Exception):
...
+            mode = 'rt'
+        else:
+            mode = 'rb'
         if self.dataset.endswith('.gz'):
-            return gzip.open(self.dataset)
+            fd = gzip.open(self.dataset, mode)
         else:
             if six.PY2:
-                return open(self.dataset, 'rb')
+                fd = open(self.dataset, mode)
```
Axik: I don't see difference between six.PY2 and PY3, they looks the same:
`fd = open(self.dataset, mode)`
and
`fd = open(self.dataset, mode)`

Axik: it could be just
```
else:
   fd = open(self.dataset, mode)
return fd
```

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -101,65 +178,105 @@ def __iter__(self):
...
-                                                      self.chunksize)):
-            if batch_num == 0:
-                self._ui.debug('input head: {}'.format(pformat(chunk[:2])))
-
-            yield Batch(rows_read, fieldnames, chunk, self.rty_cnt)
-            rows_read += len(chunk)
-        if batch_num is None:
-            raise ValueError("Input file '{}' is empty.".format(self.dataset))
-
+        if self.fast_mode:
```
Axik: can we move this validation in FastReader?

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -101,65 +178,105 @@ def __iter__(self):
...
 
     def next(self):
         return self.__next__()
 
     def push(self, batch):
         # we retry a batch - decrement retry counter
-        with self.lock:
-            batch = batch._replace(rty_cnt=batch.rty_cnt - 1)
-            self.deque.append(batch)
+        batch = batch._replace(rty_cnt=batch.rty_cnt - 1)
```
Axik: should we do it with lock?

datarobot_batch_scoring/batch_scoring.py
Axik: or at least something like in
`is_multiline`

datarobot_batch_scoring/network.py
```diff
@@ -27,7 +27,10 @@ def _request(self, request):
             response = session.send(prepared, timeout=self._timeout)
         except Exception as exc:
             logger.warning('Exception {}: {}'.format(type(exc), exc))
-            callback = request.kwargs['hooks']['response']
+            try:
+                callback = request.kwargs['hooks']['response']
+            except AttributeError:
```
Axik: what is for?

datarobot_batch_scoring/network.py
```diff
@@ -27,7 +27,10 @@ def _request(self, request):
...
+                callback = request.hooks['response'][0]
```
Axik: what is this case?

Axik:
Looks good to me except some notes.

datarobot_batch_scoring/batch_scoring.py
pprett: reduce is used in the Context class

datarobot_batch_scoring/__init__.py
pprett: pumped to 1.7.0a0

Makefile
pprett: thanks! i missed that -- i sued it during testing

datarobot_batch_scoring/batch_scoring.py
pprett: thanks

datarobot_batch_scoring/batch_scoring.py
pprett: good point -- moved it to `FastReader`

Axik:
Also, do we want to have --dry-run option in _release_?

pprett:
I personally find it helpful (e.g. to test throughput of your hdd) but no
hard feelings if we decide to drop it.

On Mon, May 23, 2016 at 11:20 AM, Kyrylo Perevozchikov <
notifications@github.com> wrote:

> Also, do we want to have --dry-run option in *release*?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/datarobot/batch-scoring/pull/25#issuecomment-220929215>
>



-- 
Peter Prettenhofer
DataRobot Inc.

pprett:
got a LGTM from @Axik via phone

PR #30, opened 2016-06-10
Couple optimisations

nimnull:
Iterable usage. Fixed session spawning. Added hack for freeing unused objects after making request


mckaymatt:
The changes to network.py seem to keep memory usage constant.

Axik:
LGTM

datarobot_batch_scoring/network.py
```diff
@@ -19,12 +21,12 @@ class Network(object):
     def __init__(self, concurrency, timeout):
         self._executor = ThreadPoolExecutor(concurrency)
         self._timeout = timeout
+        self.session = requests.Session()
```
pprett: we might watch out here -- Session appears to make no statements on thread-safety 

just found this note: http://stackoverflow.com/questions/18188044/is-the-session-object-from-pythons-requests-library-thread-safe

datarobot_batch_scoring/network.py
```diff
@@ -33,6 +35,12 @@ def _request(self, request):
                 callback = request.hooks['response'][0]
             response = FakeResponse(400, 'No Response')
             callback(response)
+        finally:
+            request.data = None
+            prepared.data = None
+            del request
+            del prepared
+            gc.collect()
```
pprett: this is quite costly after every request -- please note that the process running the thread pool is performance critical

furthermore, i can still see memory leakage... i suppose we either have circular references somewhere in our requests callback.

Axik: How the leakage looks like? Btw - one of the things to switch to threading was that 16 number of workers allocated gigantic amount of memory. Otherwise I suppose we want to investigate it. Is it so critical? Batch-scoring wasn't priority for a long time.

PR #32, opened 2016-06-16
Auto sampler and Windows bugfix

mckaymatt:
From benchmark testing I found that batch-scoring throughput really depended on using the proper batch size, and that the proper size really depended on the width dataset. This PR adds a function that finds out how many rows are needed to reach a batch of about 1.5MB. 

This PR also attempts to address an issue on windows caused by problems pickling the csv.dialect object. It does this by registering the dialects internally in the way that the python documentation recommends. This is worth doing even if it doesn't fix the Windows issue.

I also refactored encoding and dialect detection out into utils.py


smillet15:
It seems like this should really be 2 distinct PRs.

PR #33, opened 2016-06-16
Use the built-in CSV module registry to store dialect.

mckaymatt:
This is an attempt to fix an issue with Windows where Python errors while pickling the dialect object. The script now uses the csv modules built in registry for dialects.

I also refactored  _investigate_encoding_and_dialect_  and _Recoder_ out into the utils module.


mckaymatt:
@datarobot/predictions @smillet15

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -520,13 +443,13 @@ def __init__(self, n_samples, out_file, pid, lid, keep_cols,
         self.file_context = file_context
         self.fast_mode = fast_mode
         self.encoding = encoding
-        self.dialect = dialect
+        self.dialect = csv.get_dialect('dataset_dialect')
```
nimnull: If this call will preface `investigate_encoding_and_dialect()` then it will ends up in raised `Error`. Is this anyway possible? 
Also, it is little bit mysterious for outside spectator how the dialect named `dataset_dialect` appears. Should it be commented at least?

mckaymatt: Sure.

Axik:
LGTM

PR #37, opened 2016-06-23
Don't confuse user with fakeresponse

Axik:
When requests raising ReadTimeoutError, we're crafting FakeResponse and
processing it. Instead of processing it's like 400 response we should
raise information that batch was dropped because server was responding
too long


Axik:
@mckaymatt

PR #42, opened 2016-07-11
make manager NT only

mckaymatt:
The manager seems to have a significant impact on performance; dry_run takes about 30% longer.  Let's disable it when we are not running on Windows.


datarobot_batch_scoring/batch_scoring.py
```diff
@@ -742,11 +742,18 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
     t1 = time()
     queue_size = concurrent * 2
     with ExitStack() as stack:
-        manager = stack.enter_context(multiprocessing.Manager())
-        queue = manager.Queue(queue_size)
-        deque = manager.Queue(queue_size)
-        lock = manager.Lock()
-        rlock = manager.RLock()
+        if os.name is 'nt':
```
Axik:
```
if os.name is 'nt':
      conc_manager = stack.enter_context(multiprocessing.Manager())
else:
      conc_manager = multiprocessing
queue = conc_manager.Queue(queue_size)
deque = conc_manager.Queue(queue_size)
lock = conc_manager.Lock()
rlock = conc_manager.RLock()
```

what about something like this

Axik:
Lgtm on green

PR #43, opened 2016-07-28
Allow users to skip encoding and dialect detect

mckaymatt:
add option to define document encoding
add option to skip csv dialect detection.
make adjustment to sample size used by dialect and encoding detection
use auto_sample as default unless "--n_samples" is defined
allow "tab" command line arg keyword. e.g. "--delimiter=tab"


mckaymatt:
@datarobot/predictions  Please review.

errno:
LGTM

PR #44, opened 2016-08-23
improve url parsing, client error logging, include version in user-agent

mckaymatt:
-  improve url parsing with `parse_host` func and log problems faster
-  improve ambiguous client errors messages in the `authorize` func. Tell the user if they were redirected, and add specific messages for 403s and 502s. 
-  add the version of the batch-scoring script in the user-agent header


.gitignore
```diff
@@ -11,3 +11,4 @@ build
 .install-deps
 .shelve
 htmlcov
+.idea
```
dchudz: don't want this, right?

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -677,6 +679,23 @@ def batch_generator_args(self):
...
+    Log a warning to the user if it was redirected
+    """
+    history = req.history
+    if history:
+        first = history[0]
+        if first.is_redirect:
+            starting_endpoint = first.url  # Requested url
+            redirect_endpoint = first.headers.get('Location')  # redirect
+            if str(starting_endpoint) != str(redirect_endpoint):
+                ui.warning('The requested URL:\n\t{}\n\twas redirected '
```
dchudz: is redirecting a common/expected occurence? when does/would it happen?

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -677,6 +679,23 @@ def batch_generator_args(self):
         return args
 
 
+def is_redirect(req, ui):
```
dchudz: I think `warn_if_redirect` would be a better name for this. `is_redirect` sounds like it returns a boolean (whereas your function is called only for the side effects)

datarobot_batch_scoring/batch_scoring.py
mckaymatt: It usually means they used http instead of https.  The function only runs when there is already a problem and it would almost certainly help to know if they were redirected.

dchudz: ah, cool

mckaymatt: I didn't include tests because this function just logs. It would be a lot of work to simulate this in our fixtures and I don't know if it would have many benefits.

dchudz: yeah, not complaining about the lack of tests

mckaymatt:
@datarobot/predictions

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -741,6 +776,15 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
     multiprocessing.freeze_support()
     t1 = time()
     queue_size = concurrent * 2
+    #  provide version info and system info in user-agent
+    base_headers['User-Agent'] = 'datarobot_batch_scoring/{}|' \
```
Axik: This is great idea.

Axik:
We need to reflect in docs of changes that now `/api` addin by default.

CHANGES.rst
```diff
@@ -1,3 +1,11 @@
+1.8.6 (2016 August 23)
+==================
+* improve url parsing
```
Axik: Can you elaborate on this, like _you don't need to include `api` to your host variable anymore_

Axik:
LGTM, let's merge it on green.

PR #45, opened 2016-10-25
Split big requests to fit server-side limits

sevikkk:


dchudz:
Is the dockerfile stuff really about `Split big requests to fit server-side limits`? Or is this PR combining two separate things?

sevikkk:
Really it's 3 separate things:

splitting of requests
increasing auto-sampler target size
dockerizing of thing

but last two are really small ones...

dchudz:
I find that code review is easier (and higher quality) when we keep separate things separate... and it's nice for thing 1 to not be blocked by concerns about thing 2, etc.

PR #46, opened 2016-10-25
Files to run batch-scoring in docker

sevikkk:


dchudz:
What's the reason for dockerizing here? 
Does it help us run the tests? 
Has the docker image been pushed to the DataRobot docker registry?
Should the package include a mechanism (e.g. makefile target) for building/pushing the image?

sevikkk:
I did development of this thing in Docker using docker-compose and these files.

It is not necessary to distribute images somewhere to use Docker you can build and use them locally. And docker-compose way to use is just ```docker-compose up``` in top level directory.

Dockerfile-2.6
```diff
@@ -0,0 +1,32 @@
...
+
+WORKDIR /opt/project
+
+RUN grep -v '^-e .' requirements.txt >/tmp/requirements.txt && \
+	pip install -r /tmp/requirements.txt && \
+	rm /tmp/requirements.txt
+
+COPY . /opt/project
+
+RUN pip install -e .
```
nimnull: does this command do the same as the requirements installation above?

requirements.txt
```diff
@@ -7,4 +7,9 @@ pymongo==2.8
 Flask==0.10.1
```
nimnull: this file should be discarded.
We can use `setup.py` to track all third-party libs this package requires.

Dockerfile-2.6
sevikkk: I'm filtering out "-e ." in first step of dependencies install - so we don't need to reinstall them each time

This way rebuild after sources change is mostly cached

requirements.txt
sevikkk: And we have some additional requirements like flakes8 to run tests, that not needed for package itself, so they are not present in setup.py

nimnull: ok, it works, but then docker become an extra requirement for development. It sounds really weird. It expect extra work to integrate it with debugging tools. I'm not sure it is a good way to go.

Axik:
major improvement 👍 

One question - why ubuntu-trusty and not alpine?

sevikkk:
It is losely based on https://hub.docker.com/r/shimizukawa/python-all/ image

nimnull:
I'd better recommend to use these bases:
https://github.com/datarobot/DataRobot/tree/master/dockerfiles/python

requirements.txt
sevikkk: Why docker s required? You can use it with virtualenv/etc as usual

if requirements.txt will be read into into setup.py it will work without docker as well... and of course it will work if we exctract initial set of requirements from setup.py some way

sevikkk:
there is no supported 2.6 base image :) for 2.7+ these bases are used

smillet15:
@nimnull This is an open source project. I don't think it should use an internal DataRobot base image at all.

sevikkk:
sorry, i've missed datarobot part, these images are based on dockerhub images

smillet15:
@sevikkk Right. As they should be for this project.

nimnull:
@smillet15 agree, but officially supported images (that also resides on dockerhub) seems more reasonable thing to use instead of single maintainer's builds

sevikkk:
there is 3 Dockerfiles currently in this project:
Dockerfile-2.6 - based on official ubuntu:trusty image and uses ppa:deadsnakes for python2.6 debs
Dockerfile-2.7 - based on official python:2.7 image
Dockerfile-3.5 - based on official python:3.5 image

None of them is using single-maintainer builds :)

Axik:
Fine by me. I also think dockerized batch_scoring could be used in test-any-predictions approach.

PR #48, opened 2016-10-25
Split big requests until they fit to server-side limits

sevikkk:


tests/liveserver_fixtures.py
```diff
@@ -152,7 +153,23 @@ def post_auth():
 
     @app.route('/api/v1/<pid>/<lid>/predict', methods=["POST"])
     def predict_sinc(pid, lid):
+        body = flask.request.data
+        body = body.decode('utf-8').splitlines()
+        rows = len(body) - 1
+
+        try:
+            # for files with row_id as first column
+            first_row = int(body[1].split(',')[0]) % 10
```
dchudz: I don't think I understand this part

dchudz:
how does the user control the `max_batch_size`? does this need to be added to the (already long) list of command line options?

someone else should code review, since I've never worked on batch scoring

sevikkk:
max_batch_size is really not an user-controllable thing, it's request body size limit on our backend servers. So this thing in not supposed to be triggered often, it's just a safety mechanism to prevent errors on service usage. It's why we have auto-sampler at 2.5M - so in most cases this thing will not be triggered and we will not get performance penalty of multiple encoding tries.

tests/liveserver_fixtures.py
sevikkk: before this thing was always called with 10 requests long batch and was answering with static reply.

To check splitting logic, I have set max_batch_size to pretty small value (100 bytes) in test_regression tests, and now this code is called with 10 and 5 records batches, so I had to modify this endpoint to properly return top or bottom half of static output for 5 rows requests.

sevikkk: in body of request to this endpoint we have csv file with header and row_id in first column^ so this line just extracts starting_row_id of incoming dataset.

This endpoint is used in some other tests as well, and these places do not follow this pattern, so for them we return just complete static entry as before

mckaymatt:
@dchudz  the batch sizes are designed to maximize throughput. 2 to 2.5 MB seems (based on trial and error) to be a good number. Above that size you don't see any performance gains but below about 1MB you do start seeing a penalty.

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -412,23 +416,53 @@ def __iter__(self):
...
+                                                        len(data)))
+
+                    if batch.rows < 2:
+                        self._ui.error('batch {} is single row but bigger '
+                                       'than limit, skipping, we lost {} '
+                                       'records'.format(batch.id,
+                                                        len(batch.data)))
+                        continue
+
+                    split_point = int(batch.rows/2)
```
mckaymatt: We already form predefined chunks here https://github.com/datarobot/batch-scoring/blob/master/datarobot_batch_scoring/utils.py#L293  I'm not sure we should replicate the same logic in another place.

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -412,23 +416,53 @@ def __iter__(self):
...
-                url=self.endpoint,
-                headers=self.headers,
-                data=data,
-                auth=(self.user, self.api_token),
-                hooks={'response': hook})
+            todo = [batch]
+            while todo:
+                batch = todo.pop(0)
+                data = chunk_formatter(batch.data, batch.fieldnames)
+                if len(data) < self.max_batch_size:
```
mckaymatt: Most of the necessary logic exists in this function. https://github.com/datarobot/batch-scoring/blob/master/datarobot_batch_scoring/utils.py#L420 We could probably a test for max-size here

dchudz: Do we need to worry about e.g. text features, where different rows might have very different sizes?

What's the outcome if we are wrong and a future batch is too big? Is that a bad situation or okay?

I'm not too familiar with batch_scoring, so just asking.

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -772,7 +807,8 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
                           out_file, keep_cols, delimiter,
                           dataset, pred_name,
                           timeout, ui, fast_mode, auto_sample,
-                          dry_run, encoding, skip_dialect):
+                          dry_run, encoding, skip_dialect,
+                          max_batch_size=3.5 * 1024 ** 2):
```
dchudz: I'm not sure I follow all of the logic of how these parameters are passed through batch_scoring, but... are we essentially using a default parameter value as a constant here? That's not so good.

datarobot_batch_scoring/batch_scoring.py
mckaymatt: It wouldn't be much of a problem if we are sending larger or smaller batches. That's why I prefer an approximate solution.

datarobot_batch_scoring/batch_scoring.py
sevikkk: We have assumption on nearly equivalent size of each rows in autosampler, but looks like in some cases this assumption is broken when we have some not so regular datasets.

So it was implied that this code needs to check each chunk to ensure we are not hitting limits.

datarobot_batch_scoring/batch_scoring.py
dchudz: I thought the server is going to reject batches higher than 10mb?

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -412,23 +416,53 @@ def __iter__(self):
...
+                        headers=self.headers,
+                        data=data,
+                        auth=(self.user, self.api_token),
+                        hooks={'response': hook})
+                else:
+                    self._ui.debug('batch {}-{} is too long: {} bytes,'
+                                   ' splitting' .format(batch.id, batch.rows,
+                                                        len(data)))
+
+                    if batch.rows < 2:
```
sevikkk: it's size of data section (number of actual data rows) before formatting and putting header in place.

And we can get here after log2(n) splits in case we have single row with size bigger than limit.

datarobot_batch_scoring/batch_scoring.py
mckaymatt: I think the overall goal is to keep people from trying to send 500MB chunks. I'm not sure we need to be that precisely about batch size.

dchudz: anything above the limit (generally 10mb) will be an error, right?

datarobot_batch_scoring/batch_scoring.py
sevikkk: Server will reject batches larger than 10mb, and this thing tries to prevent exactly this thing. We are not trying to format all batches to exactly X megs, we just check final size of encoded batch with 0 overhead, and do something only if it's size bigger than limit.

It's not supposed to happen often

dchudz: makes sense to me. except it looks like we're hard-coding the 10mb limit as a default function value? which I'm skeptical of b/c:

(1) implementing constants as default values is bad
(2) this is a server config that could change, so users may need to change it

sevikkk: i'm following style in auto-sampler :)
I can fix both of them of course

dchudz: auto-sampler is maybe a different story, if user doesn't *need* control over it? I'm thinking if the 10mb parameter for the server ever changes, user would need to control this in the client too?

but I don't know anything about batch_scoring so I could be totally wrong. @mckaymatt or @Axik or someone would probably know best.

sevikkk: But auto-sampler limit is based on performance model of the current implementation of prediction and most probably will change at the same time as body size limit :)

dchudz: @sevikkk true, but I figured non-ideal performance is more okay than errors due to too-large batches :)

mckaymatt: I didn't realize we had a 10M limit. That does change things.

datarobot_batch_scoring/batch_scoring.py
mckaymatt: I agree that this doesn't need to be another argument to the main function.  There are enough of those already.
My initial comment was more aimed at the fact that we are setting max size to "3.5 * 1024 ** 2" instead of "10 * 1024 ** 2". I'm confused by that.

datarobot_batch_scoring/batch_scoring.py
mckaymatt: Gotcha. I see why it's there, but I just didn't understand why the block was needed earlier. Didn't realize we had a 10M limit.

datarobot_batch_scoring/batch_scoring.py
sevikkk: putting it as function argument was needed mainly for tests, so I can set it low and avoid 10m+ file just to test it will be split properly.

mckaymatt: That's fair but is there any way around it? Can we patch it in the tests? Can we used a giant, very repetitive gzipped file?

sevikkk: It can be replaced with module level global patched during tests, but i'm not sure it has a sense. It's just like syntetic time passed as parameter to functions to avoid using real-time clock to allow testability...

sevikkk: So from my point of view putting in some things for better testability is not bad at all.

datarobot_batch_scoring/batch_scoring.py
mckaymatt: Should we warn or fail the script at the beginning we estimate that the --n_samples will be far more than 10MB?  Maybe that's an edge case since we really want people to use auto_sample now

datarobot_batch_scoring/batch_scoring.py
Axik: The only thing  - it's still should be a constant variable that defined in the beginning of module. Otherwise - it's good to pass it explicitly.

dchudz: `it's still should be a constant variable that defined in the beginning of module`

exactly

Axik:
LGTM

PR #51, opened 2016-11-01
Fix memory leak

mckaymatt:
The problem with using executor.map is that it was greedy and it would keep calling __next__() on the WorkUnitGenerator even when every thread was busy. 
With this fix we do not call __next__() until a worker thread is finished.


mckaymatt:
I also removed the manual garbage collection since it didn't seem like it made any difference during my manual testing.

mckaymatt:
@datarobot/predictions  Please review

datarobot_batch_scoring/network.py
```diff
@@ -45,15 +46,23 @@ def _request(self, request):
...
-            del prepared
-            gc.collect()
 
     def perform_requests(self, requests):
-        return self._executor.map(self._request, requests)
+        for r in requests:
+            while True:
+                self.futures = [i for i in self.futures if not i.done()]
+                if len(self.futures) < self.concurrency:
+                    self.futures.append(self._executor.submit(self._request,
```
pprett: my bad - the break below

mckaymatt: We are iterating the WorkUnitGenerator sequentially. The batches generated by the WorkUnitGenerator are then being passed to the executor where they are being executed in threads non-sequentially. The only difference between this approach and the previous approach is that we do not submit additional batches until some of the outstanding threads finish executing. 
https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.submit

PR #52, opened 2016-11-02
--output_delimiter and --skip_row_id flags (PRED-908, PRED-909)

sevikkk:
Sorry, but there is no way to make them separate without conflicts on merge.

Configuration changes are the same in the same places as well.


pprett:
hmm.. CHANGES.rst became a binary file?

README.rst
```diff
@@ -84,6 +84,8 @@ The following table describes each of the arguments:
  create_api_token              Requests a new API token. To use this option, you must specify the ``password`` argument for this request (not the ``api_token`` argument). Specifying this argument invalidates your existing API token and creates and stores a new token for future prediction requests.
  n_retry=<n_retry>             Specifies the number of times DataRobot will retry if a request fails. A value of -1, the default, specifies an infinite number of retries.
  pred_name=<pred_name>         Applies a name to the prediction column of the output file. If you do not supply the argument, the column name is blank. For binary predictions assumes last class in lexical order as positive.
+ skip_row_id                   Skip the row_id column in output.
+ output_delimiter=<delimiter>  Specifies delimiter for output CSV
```
mckaymatt: Include the fact that they can use the word "tab". see line 90

datarobot_batch_scoring/utils.py
```diff
@@ -410,7 +413,8 @@ def investigate_encoding_and_dialect(dataset, sep, ui, fast=False,
     csv.register_dialect('dataset_dialect', dialect)
     #  the csv writer should use the systems newline char
     csv.register_dialect('writer_dialect', dialect,
-                         lineterminator=os.linesep)
+                         lineterminator=os.linesep,
+                         delimiter=(output_delimiter or dialect.delimiter))
```
mckaymatt: in python 2 we might need to call str() on this. If you give it Unicode the csv parser complains.

sevikkk: it's already fixed for autodetected delimiter, and one comming from commandline already checked against of set of allowed delimiters..

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -569,10 +577,19 @@ def checkpoint_batch(self, batch, out_fields, pred):
                     # FIXME this will fail on quoted fields!
                     row = row.rstrip().split(input_delimiter)
                 keeps = [row[i] for i in indices]
-                comb.append([predicted[0]] + keeps + predicted[1:])
+                output_row = []
```
Axik: we already have a `comb`. No need for output_row

sevikkk: comb is list of all rows, not single current row

datarobot_batch_scoring/utils.py
mckaymatt: In Python 2 a  u'A'  == "A",  so that won't necessarily work. Just make sure str() gets called on `(output_delimiter or dialect.delimiter))` if it's not None. 
Unicode will be rejected by the csv module

mckaymatt: I think validating that the output_delimiter is a bytestring in Python2 is a necessary change.

PR #54, opened 2016-11-03
Track prediction errors, warnings and total stats (PRED-910)

sevikkk:
Added tracking of error and warning messages

Added warnings, errors and stats sections at the end of script output


datarobot_batch_scoring/utils.py
```diff
@@ -410,7 +413,8 @@ def investigate_encoding_and_dialect(dataset, sep, ui, fast=False,
     csv.register_dialect('dataset_dialect', dialect)
     #  the csv writer should use the systems newline char
     csv.register_dialect('writer_dialect', dialect,
-                         lineterminator=os.linesep)
+                         lineterminator=os.linesep,
+                         delimiter=str(output_delimiter or dialect.delimiter))
```
mckaymatt: What if output_delimiter is type None?

sevikkk: dialect.delimiter can't be None

mckaymatt: We should test with the --skip_dialect flag

sevikkk: --skip-dialect forces Excel dialect Ok, i'll add test :)

sevikkk:
@pprett

PR #55, opened 2016-11-04
move response processing into seperate process

mckaymatt:
The main thing slowing down the batch-scoring script is that we are performing network IO and CPU bound post-processing in the same process. These changes move the CPU bound post-processing into a separate process.

Not all of the plumbing is hooked up yet, tests will  fail and this wouldn't work in Windows;  but you can score with this commit. 
Using the 1.8.7 branch I scored a 500M file in 75 seconds. 
Using this commit on the same file it scored in 41 seconds.
That's almost double the speed  :chart_with_upwards_trend:  This should def be in the next release


mckaymatt:
cc @datarobot/predictions

PR #56, opened 2016-11-08
see if windows tests pass now

mckaymatt:


mckaymatt:
We were skipping some tests because they would fail on Windows with Python3.5 in Appveyor. I suspected that executor.map was causing the problems and wanted to see if they would pass now... and it looks like they do. We could fully test Windows now.

The only real issue is that the Windows tests on Appveyor take forever. 

Thoughts? @sevikkk @Axik

sevikkk:
Looks like Appveyor is especially slow on tests with big fixtures like regression_jp and regression_bad. But files are just 200k...

sevikkk:
With changes to n_samples tests run much faster

mckaymatt:
Oh wow, that's a huge improvement. I'm comfortable merging this.

sevikkk:
I'm comfortable too, it's your PR :)

mckaymatt:
Oh I need to update the docs

PR #58, opened 2016-11-21
Set connection pool size according to concurrency (PRED-936)

sevikkk:


tests/test_network.py
```diff
@@ -52,3 +53,48 @@ def test_request_client_timeout(live_server, tmpdir):
 or
 increase "--timeout" parameter.
 """))
+
+
+def test_request_pool_is_full(live_server, tmpdir, monkeypatch):
```
Axik: is this test not passing without fix?

Axik: Nice

PR #59, opened 2016-11-24
Matt/fastwrite rebase

mckaymatt:
There is still a lot of work to do on this. I haven't run tests yet but it works.


datarobot_batch_scoring/batch_scoring.py
```diff
@@ -303,6 +300,8 @@ def _shove(self, args, dialect, queue):
                 _ui.debug('queueing batch {}'.format(batch.id))
                 queue.put(batch)
 
+            _ui.info('shoveling complete | total time elapsed {}s'
```
mckaymatt: This log statement had been in the wrong place.

tests/liveserver_fixtures.py
```diff
@@ -167,6 +169,15 @@ def predict_sinc(pid, lid):
         except ValueError:
             first_row = 0
 
+        # This is to support testing retries. It's Maybe a little hacky...
+        # look up the "pids" to see where this logic comes into play
+        if app.count < 5 and pid == 'five_tries_f7ef75d75f164':
```
mckaymatt: This might be terrible but it works and now we test the deque

mckaymatt:
Please review @datarobot/predictions

sevikkk:
For faster turn-around we probably need to run test-cases with disabled fork's on linux as well:

 https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method

So we will catch problems with inherited resources on WIndows faster...

datarobot_batch_scoring/__init__.py
```diff
@@ -1 +1 @@
-__version__ = '1.8.8'
+__version__ = '1.9.0'
```
Axik: I thought about releasing 1.9.0 with refactoring and compression also.

mckaymatt: I figure this would be a good place to do a minor bump because these changes are more likely to introduce instability. We can always say "go back to 1.8.x "  if there are issues.

mckaymatt: or go back to 1.8 if this runs like crap on a 2 core system. Our "script" might need to include system requirements now.

mckaymatt:
There is some issue on Windows Py27 with pickling staticmethods. It's looking at `datarobot_batch_scoring.batch_scoring.run_subproc_cls_inst` instead of 
`datarobot_batch_scoring.batch_scoring.WriterProcess.run_subproc_cls_inst`.

I turned it into a regular function and it seems to work now.

mckaymatt:
Any more comments?

I have a branch, based on this on, for compression.  https://github.com/datarobot/batch-scoring/pull/61

PR #60, opened 2016-11-24
Refactoring in preparation to split into multiple workers

sevikkk:


datarobot_batch_scoring/network.py
```diff
@@ -62,7 +263,7 @@ def perform_requests(self, requests):
                                                               r))
                     break
                 else:
-                    sleep(0.1)
+                    wait(self.futures, return_when=FIRST_COMPLETED)
```
mckaymatt: Nice.

PR #61, opened 2016-12-01
add option to compress data

mckaymatt:
I have to head out so I don't have time to fix all the tests, but this does work against Kyrylo's DR branch.  

A few notes:
I couldn't get this to work using zlib, so I used gzip. The server would complain and say it wasn't the right format when I used zlib.  
We might want to move the compression into the shovel process... but that's debatable.

@Axik @sevikkk


sevikkk:
Tests are fixed and some disabled ones for known problems added

Known ones are out of scope for this PR

PR #63, opened 2016-12-01
use a lower default compression

mckaymatt:
This compresslevel has a more reasonable trade off between time and space savings. We also now log space savings instead of compression ratio.

I also clean up context if it's a dry_run

@Axik @sevikkk

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@
 * added 3rd parallel process which handles post-processing and writing of responses.
   This should greatly improve performance.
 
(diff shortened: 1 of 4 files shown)


mckaymatt:
if success is true it allows the runcontext to remove the .shelve file when it exits. We should clean up the shelve if we succeed.

mckaymatt:
Our README doesn't render properly on pypi. I fixed the markdown and added validation to our `test` target
https://pypi.python.org/pypi/datarobot_batch_scoring/1.8.8

PR #64, opened 2016-12-07
batch-scoring multi-process workers

sevikkk:


mckaymatt:
This is what the output looks like on windows. When you run it you can't really tell if it's working
```
MainProcess [INFO] platform: win32 2.7.12 |Anaconda 4.2.0 (64-bit)| (default, Ju
n 29 2016, 11:07:13) [MSC v.1500 64 bit (AMD64)]
MainProcess [INFO] connecting to https://xxxxx
com/api/v1/
MainProcess [INFO] auto_sampler: total time seconds - 0.0699999332428
MainProcess [INFO] auto_sample: will use batches of 14980 rows
MainProcess [INFO] Shovel go...
MainProcess [INFO] Writer go...
MainProcess [INFO] Network go...
MainProcess [INFO] shovel proc finished, exit code: 0
MainProcess [INFO] All requests done, waiting for writer
MainProcess [INFO] network proc finished, exit code: 0
MainProcess [INFO] writer proc finished, exit code: 0
MainProcess [INFO] all workers exited successfully
MainProcess [INFO] Shovel is finished ok. Chunks produced: 409 Resource usage: {
}
MainProcess [INFO] Network is finished ok. Chunks: 409 Requests: 409 Retries: 0
Resource usage: {}
MainProcess [INFO] Writer is finished ok. Result: True Results: 409 Written: 409
 Resource usage: {}
MainProcess [INFO] scoring complete | time elapsed 188.612999916s
MainProcess [INFO] scoring complete | total time elapsed 190.625s
MainProcess [INFO] ==== Scoring warnings ====
MainProcess [INFO] ==== Scoring errors ====
MainProcess [INFO] ==== Total stats ====
MainProcess [INFO] done: 6122178 lost: 0
```

datarobot_batch_scoring/writer.py
```diff
@@ -410,9 +405,45 @@ def format_result_data(self, result, batch, pred_name):
...
+            if not skip_row_id:
+                comb = pred
+                written_fields = out_fields
+            else:
+                written_fields = out_fields[1:]
+                comb = [row[1:] for row in pred]
+
+        return (written_fields, comb)
 
     def sigterm_handler(self, _signo, _stack_frame):
```
mckaymatt: We could possibly put this in the reader. I'm not sure if it's necessary though.

datarobot_batch_scoring/network.py
```diff
@@ -218,66 +286,110 @@ def __iter__(self):
...
+            "rusage": get_rusage(),
+        }))
+
+    def go(self, dry_run=False):
+        if dry_run:
+            return self.run(dry_run)
+
+        self.ui.set_next_UI_name('network')
+        self.proc = \
+            multiprocessing.Process(target=self.run,
```
mckaymatt: Does this need to be a separate process?

datarobot_batch_scoring/writer.py
mckaymatt: maybe it would make more sense in the main process so it can gracefully shutdown the other processes.

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -70,14 +75,13 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
...
-            if not pwd:
-                pwd = ui.getpass()
-            try:
-                api_token = acquire_api_token(base_url, base_headers, user,
-                                              pwd, create_api_token, ui)
-            except Exception as e:
-                ui.fatal(str(e))
+        progress_queue = conc_manager.Queue()
+
+        shovel_status = conc_manager.Value('c', b'-')
```
mckaymatt: using this is a good idea. By default `lock` is true. If we don't actually need that lock than we should set it false.

datarobot_batch_scoring/network.py
```diff
@@ -218,66 +274,110 @@ def __iter__(self):
...
-                if len(self.futures) < self.concurrency:
-                    self.futures.append(self._executor.submit(self._request,
-                                                              r))
-                    break
-                else:
-                    wait(self.futures, return_when=FIRST_COMPLETED)
-            yield
+    def request_cb(self, f):
+        futures = [i for i in self.futures if not i.done()]
+        self.ui.debug('cb {}: pending futures: {}'.format(f, len(futures)))
```
sevikkk: cb - callback, it's native callback of future, that is called when future is actually done.

half-baked-not-so-async-ready handlers of requests library are run in target thread, before future is considered done. So we can't check number of pending futures in it.

So to get reliable detect of last one, we need to put it in real callback of ThreadPool's future.

datarobot_batch_scoring/network.py
sevikkk: As a next step I'll add a flag to run multiple Network workers, so we can run cpu-bound compression/decompression in parallel.

datarobot_batch_scoring/batch_scoring.py
sevikkk: Yes, we have just one writer, so lock can be safely disabled

datarobot_batch_scoring/utils.py
```diff
@@ -419,3 +419,23 @@ def get_rusage():
...
+    def __init__(self, status_value):
+        self.status_value = status_value
+
+    @property
+    def state(self):
+        return self.status_value.value
+
+    @state.setter
+    def state(self, status):
+        self.ui.debug('state: {} -> {}'
```
mckaymatt: Can we get the human readable format here?

datarobot_batch_scoring/network.py
mckaymatt: I timed it before and it didn't seem like compression/decompression added that much overhead. I think that might be premature optimization. The  added queue might add more overhead than compression and it makes logging on windows more complicated.

sevikkk: From Kiev office scoring on ORM instance with compression gets 130 vs 160 seconds.
So I'm not so sure it does not give much to us..

datarobot_batch_scoring/utils.py
sevikkk: done

datarobot_batch_scoring/network.py
mckaymatt: I think you misunderstood what I was saying. I was trying to say is that I do not think that we need a new process just for compression...yet. I don't think the overhead for compression is that large.  I will leave it up to you.

The resource info (which is very cool!) does show we spend a not-insignificant amount of time compressing. This is on a 1G dataset with a fast connection, 31 workers and a fast model. The bottleneck is the writer in this case. 
```
# NO COMPRESS 
MainProcess [INFO] Shovel is finished ok. Chunks produced: 409 Resource usage:                   { 'stime': 2.499195, 'utime': 12.081455, 'rss': 261308}
MainProcess [INFO] Network is finished ok. Chunks: 409 Requests: 409 Retries: 0 Resource usage:  { 'stime': 5.085765, 'utime': 23.361974, 'rss': 704068,}
MainProcess [INFO] Writer is finished ok. Result: True Results: 409 Written: 409 Resource usage: { 'stime': 1.429178, 'utime': 52.195045, 'rss': 69672,}
MainProcess [INFO] scoring complete | time elapsed 61.48334240913391s
MainProcess [INFO] scoring complete | total time elapsed 61.92390727996826s

# WITH COMPRESS 
MainProcess [INFO] Shovel is finished ok. Chunks produced: 409 Resource usage:                   {'stime': 2.523238, 'utime': 11.233651, 'rss': 260676}
MainProcess [INFO] Network is finished ok. Chunks: 409 Requests: 409 Retries: 0 Resource usage:  {'stime': 4.524693, 'utime': 45.154097, 'rss': 650000}
MainProcess [INFO] Writer is finished ok. Result: True Results: 409 Written: 409 Resource usage: {'stime': 1.59124,  'utime': 54.765873, 'rss': 68696}
MainProcess [INFO] scoring complete | time elapsed 63.556153535842896s
MainProcess [INFO] scoring complete | total time elapsed 63.95048785209656s
```
23 vs 45 seconds.

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -117,105 +131,404 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
...
+                    ui.info("all workers exited")
+                    break
+                elif time() - phase_start > 10:
+                    ui.error("some workers are not exited, ignoring")
+                    break
+
+        if shovel_done:
+            ui.info("Shovel is finished {}. Chunks produced: {}"
+                    " Resource usage: {}"
+                    "".format(shovel_done, s_produced, s_rusage))
```
mckaymatt: s_rusage is a dict so it prints in a random order (at least between runs). Can you format it so it displays in the same order?

Bonus points if you rename the keys to something more user friendly like `SystemTime`/`UserTime`/`Memory` instead of `{'utime': 52.673207, 'rss': 68588, 'stime': 1.631294}`

datarobot_batch_scoring/writer.py
```diff
@@ -454,18 +454,10 @@ def format_result_data(self, result, batch):
                 written_fields = out_fields[1:]
                 comb = [row[1:] for row in pred]
 
-        return (written_fields, comb)
-
-    def sigterm_handler(self, _signo, _stack_frame):
```
mckaymatt: I think removing signal handling is a mistake. Without it you cannot perform a cleanup if anything besides keyboardinterupt/SIGINT is received by the master. 
If you SIGKILL the master, the workers keep going like nothing happened because the master dies immediately without any cleanup. 
If you SIGKILL/SIGINT a workers, it just dies without  triggering a program shutdown.  The program only exits much later.
Every process should have a way to catch signals and trigger a fast, but safe, shutdown.

If they SIGKILL they are on their own :)

mckaymatt: I think the workers should be able to forward the signal to the parent.

datarobot_batch_scoring/network.py
```diff
@@ -372,6 +373,16 @@ def run(self, dry_run=False):
                 self.ui.info('{} responses sent | time elapsed {}s'
                              .format(i, time() - t0))
 
+                if time() - last_report > 10:
```
mckaymatt: This is nice for the windows people. Could we do it more often? Maybe every 5 seconds?

mckaymatt: This is nice for the windows people. Could we do it more often? Maybe every 5 seconds?

datarobot_batch_scoring/writer.py
```diff
@@ -410,53 +416,108 @@ def format_result_data(self, result, batch, pred_name):
                            key=operator.itemgetter('row_id'))]
             out_fields = ['row_id', pred_name if pred_name else '']
         else:
-            ValueError('task "{}" not supported'
-                       ''.format(result['task']))
-        return (out_fields, pred)
+            raise ValueError('task "{}" not supported'
+                             ''.format(result['task']))
 
-    def sigterm_handler(self, _signo, _stack_frame):
```
mckaymatt: I think removing signal handling is a mistake. Without it you cannot perform a cleanup if anything besides keyboardinterupt/SIGINT is received by the master.
If you SIGKILL the master, the workers keep going like nothing happened because the master dies immediately without any cleanup.
If you SIGKILL/SIGINT a workers, it just dies without triggering a program shutdown. The program only exits much later.
Every process should have a way to catch signals and trigger a fast, but safe, shutdown. 

Probably workers should be able to forward the signal to the parent.

If they SIGKILL they are on their own :)

sevikkk: if you sigkill the master - nothing is going to help you.. Removing everything, but supervision from master was done exactly for this reason - master is not supposed to die in any case.

master itself is looking for all children and triggers proper shutdown if any of them die.

if you sigkill any of workers - it will trigger "worker diead without reporting success" in master and anything will be shutdown properly.

datarobot_batch_scoring/batch_scoring.py
sevikkk: Really it's system specific, in windows it's empty, on linux and mac os memory is returned in different units. So I'll prefer to treat it as internal debug-only thing.

To be user friendly it must be converted to cpu%, MBytes and so on, but i'm not so sure it has any sense..

datarobot_batch_scoring/network.py
sevikkk: Current writer is not a writer, it's formatter mostly and can be split as well. And if we are going to compress on result side - it will be even more CPU bound as well. And now with concurency ~30 on 32 core target you'll not get full utilization of host really - you can check it's cpu usage.

To get full utilization of host in current scheme we need concurrency in 2x(host CPUs) range. And you'll push into local cpu limit as well in this case.

So really we need to scale out both network and writer if we want to utilize relatively big targets.

Of course it's all depends on what real use this thing is target for :)

datarobot_batch_scoring/batch_scoring.py
mckaymatt: Can we just sort the keys so it is in the same order?

datarobot_batch_scoring/writer.py
```diff
@@ -410,63 +419,147 @@ def format_result_data(self, result, batch, pred_name):
...
+                    if result is False:  # unpack_request_object failed
+                        continue
+
+                    (written_fields, comb) = self.format_result_data(result,
+                                                                     batch)
+
+                    self.ctx.checkpoint_batch(batch, written_fields, comb)
+                    written += 1
+                    rows_done += batch.rows
+                    # self._ui.debug('Writer Queue queue length: {}'
```
mckaymatt: I think this can be removed.

sevikkk: it's mainly for windows users, where you can't easily guess current row number

mckaymatt: The commented-out code?

sevikkk: gr, It's not too helpful with pointing exact line :)
I'll remove commented out things

sevikkk: done

mckaymatt:
This seems okay to me. Still needs the changes / version updated. 
I wouldn't push the tags though until we merge the OSX testing  pr and see that those work.

PR #67, opened 2017-01-06
Added SSE endpoint build and options support

nimnull:
motivation — https://datarobot.atlassian.net/browse/PRED-832


datarobot_batch_scoring/batch_scoring.py
```diff
@@ -134,7 +137,7 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
                     [c for c in keep_cols if c not in first_row.fieldnames],
                     first_row.fieldnames))
 
-        if not dry_run:
+        if not (dry_run or import_id):
```
sevikkk: Probably it's better to move import_id in next check. To not mix dry_run processing and missing data fix...

datarobot_batch_scoring/main.py
```diff
@@ -252,12 +254,18 @@ def main(argv=sys.argv[1:]):
     skip_row_id = parsed_args['skip_row_id']
     output_delimiter = parsed_args.get('output_delimiter')
 
+    if pid and not lid:
```
nimnull: Request will fail cause there probably no such URL for standalone disabled cluster
this is how argparse works ¯\_(ツ)_/¯ even if we'll use third argument — `pid` will be filled.

datarobot_batch_scoring/main.py
```diff
@@ -252,12 +254,18 @@ def main(argv=sys.argv[1:]):
...
+    else:
+        import_id = None
+
     if not os.path.exists(parsed_args['dataset']):
         ui.fatal('file {} does not exist.'.format(parsed_args['dataset']))
 
     try:
-        verify_objectid(pid)
-        verify_objectid(lid)
+        pid and verify_objectid(pid)
```
nimnull: this verification won't trigger without both `pid` & `lid` actually.

datarobot_batch_scoring/batch_scoring.py
sevikkk: Reaaly this authorize request check is serving as early check for validity of model/project/incoming data as it's done with first row of incoming file before starting threads and so on, so it fails much earlier.

So really we need to do it for SSE as well.

nimnull: we don't use authorization for SSE. Neither user/password nor /user/token, so this part of logic could be skipped as it would be a dry-run call

datarobot_batch_scoring/main.py
Axik: ok

datarobot_batch_scoring/main.py
Axik: > Request will fail cause there probably no such URL for standalone disabled cluster

This is very confusing for a user.

DEV_README.rst
```diff
@@ -45,7 +45,7 @@ Cut a release candidate
   - tag release
   - push a tag to GitHub
 
-Travis bot runs authomated tests and publish new version on PyPI when
+Travis bot runs automated tests and publish new version on PyPI when
```
mckaymatt: Is this true? I thought we published the new version when we pushed a new tag.

datarobot_batch_scoring/batch_scoring.py
sevikkk: Actually if we don't fail here, workers will never abort, and just continue to push and fail each batch.

Name of this thing is really misleading, as it's not authorize, it's more like validate_prediction_params

datarobot_batch_scoring/main.py
mckaymatt: I agree. I would rather see something like `--scoring_id=` or  when people are using SSE.

nimnull: @mckaymatt @Axik 
I thought on conversion it to the option type. But then it'll break the notation of supplying model & project id as arguments. Also, pid & lid arguments should become optional as well. And it will change batch_scoring CLI interface completely.

Axik: We need than to add some additional effort to differentiate import_id from project_id. If this a looks-like-project-id and no model-id we need break execution and said about the invalid number of arguments (missing LID). If it's looks-like-import-id we need to throw something in UI and then execute it as SSE.

I also didn't quite get the part about the 
>break the notation of supplying model & project id as arguments.

This is _already_ breaking that notation because model_id implicitly become not argument (or not a _required argument_ as argparse want it).

DEV_README.rst
Axik: Right, it's publish the new version when we pushed a new tag but after it passed automated tests. If you'll read above it's says `- push a tag to GitHub` and then part about tests and publishing. So it's a right sentence in this context.

datarobot_batch_scoring/main.py
nimnull: @Axik and how about separate entry point like `batch_scoring_transferable` so we can keep original rules as is and add another one without braking anything?

Axik: That's was my thoughts. However I don't know how much effort that would be with state of code that we have, so that's why I didn't proposed it. I If we would use click that would be much easier... But overall I like this idea.

sevikkk: Point it to the same file and just check for argv[0]? :)

mckaymatt: I think we should be able to do something like 
```batch_scoring --host=<host> --user=<user> {<project_id> <model_id> | --import_id=<ammaidentifier> } <dataset_filepath> --datarobot_key=<datarobot_key> {--password=<pwd> | --api_token=<api_token>}```
with either [subcommands](https://docs.python.org/2.7/library/argparse.html#sub-commands)  or [mutual-exclusion](https://docs.python.org/2.7/library/argparse.html#mutual-exclusion).

The best thing to do might be to make  `main,py`  much dumber. Have it only accept the arguments and then pass them into a function like `validate`. `validate` could do all of the checking that is now performed by `main.py`. The advantage of this is that it would make it much easier to use batch_scoring as a module.

mckaymatt: Maybe we keep the current format with 3 positional args and we make a special string for PID,  for example:
```batch_scoring --host=http://localhost ... SSE  586d5807c80891461e6423c7 path/to/dataset.csv``` Where `"SSE"` is a special string that that indicates a stand-alone prediction 

That would look like ```batch_scoring --host=<host> --user=<user> {<project_id>| "SSE" } <model_id>  <dataset_filepath> --datarobot_key=<datarobot_key> {--password=<pwd> | --api_token=<api_token>}```

nimnull:
@mckaymatt @Axik so after a series of experiments I added second endpoint that has own validation rules for arguments. So we still follow original interface with dedicated servers and have different endpoint to score over standalone.

mckaymatt:
@nimnull You didn't update the README

nimnull:
@mckaymatt just waited for a final decision, so now can do that work

PR #69, opened 2017-01-16
Yehor/pred 1025 new output schema

nimnull:


nimnull:
//CC @datarobot/predictions for review

datarobot_batch_scoring/writer.py
```diff
@@ -397,30 +397,26 @@ def format_result_data(self, result, batch):
...
-                    sorted(predictions,
-                           key=operator.itemgetter('row_id'))]
-            out_fields = ['row_id', pred_name if pred_name else '']
+        single_row = result[0]
+        fields = sorted(record['label']
+                        for record in single_row['predictionValues'])
+
+        if pred_name is not None:
+            out_fields = ['row_id', pred_name]
+            if len(fields) == 2 and single_row['prediction'] in fields:
```
Axik: What's this code doing? Are you eliminating negative class from output values? Could please add comment here? This clause and action is extremely hard readable.

nimnull: # batch scoring returns only positive class for binary tasks if 
            # user specified pred_name option, so we eliminate negative 
            # result

datarobot_batch_scoring/writer.py
```diff
@@ -397,30 +397,29 @@ def format_result_data(self, result, batch):
...
-                             ''.format(result['task']))
+            out_fields = ['row_id'] + fields
+        try:
+            pred = [[record['rowId'] + batch.id] + [
+                pred_vals['value'] for pred_vals in
+                sorted(record['predictionValues'],
+                       key=operator.itemgetter('label'))
+                if pred_vals['label'] in fields]
+                    for record in sorted(result,
+                                         key=operator.itemgetter('rowId'))]
```
hughdbrown: It is really hard to make out what is going on here. For a start, the indentation totally obscures the scoping of some actions. Yet when I change it to this:
```
pred = [
   [record['rowId'] + batch.id] + [
        pred_vals['value']
        for pred_vals in sorted(record['predictionValues'], key=operator.itemgetter('label'))
        if pred_vals['label'] in fields
    ]
    for record in sorted(result, key=operator.itemgetter('rowId'))
]
```
it is really not so much better. This extracts from a list of dictionaries sorted by `rowId`, and the extracts are sorted by the internal `label`, filtered by `fields`.

As an aside, the fact that `fields` is a list makes this an O(n\*\*3) algorithm, more or less.

sevikkk: You are 3 months late :) But it's not O(n**3) its O(m*n*k) :) and n and k are constant for each model.

nimnull: And from the very beginning I just had an intention to support new prediction response structure with smaller number LOC than it was before. 
If you noticed — there less conditional switches and now on we have no class labels. So it was just brief workaround to swap with in place. 
Any speedup improvements highly welcomed.

PR #71, opened 2017-02-03
Offline Install Options 

mckaymatt:
Read the DEV_README for details on how the two offline install options are built. 

The OFFLINE_INSTALL_README is zipped up with the builds and explains how to perform the offline install.

You can see the pip-based offline install in action using docker containers with no network access with these steps:

make a build and start the container:
```
make offlinebundle

docker run --network=none --rm -it -v ~/workspace/batch-scoring:/batch-scoring centos:7 bash
``` 
Run this inside the container
```
useradd -m --shell=/bin/bash -u 1001 user 
su user 
cp -r /batch-scoring/dist/offlinebundle ~/
cd ~/offlinebundle
python get-pip.py --user --no-index --find-links=helper_packages/
echo 'export PATH=~/.local/bin/:$PATH' >> ~/.bashrc
source ~/.bashrc
pip install --user --no-index --find-links=helper_packages/ helper_packages/*   #    Note these must all be wheels or it will fail
pip install --user --no-index --find-links=required_packages/ required_packages/*     # don't need to be wheels
batch_scoring --version
```

The pip-based offline install works but it is probably not optimal. I don't claim to understand everything about Python packaging and so there is probably room for improvement. I'm mainly concerned about compatibility between platforms because of our reliance on wheels.   Does anyone know a better way to bootstrap a non-privileged? 

The fall-back for when the pip-based install approach fails is the single-file-executable made by PyInstalled. This works pretty well for a stand-alone program like batch-scoring, and it is very simple for the user, but it's not a solution for a lib like the public-api package. 

You can try the single-file executable in docker with:
```
make pyinstaller
docker run --network=none --rm -it -v ~/workspace/batch-scoring:/batch-scoring centos:7 bash
``` 
inside the container
```
useradd -m --shell=/bin/bash -u 1001 user 
su user 
cp -r /batch-scoring/dist/datarobot_batch_scoring_1.10.0_executables ~/
cd ~/datarobot_batch_scoring_1.10.0_executables
./batch_scoring --help
./batch_scoring_sse --help
```

I made the bundles using `zip` for the sake of Windows users, but it doesn't seem like `unzip` comes by default on most Linux distros. I'm not sure what the most universal option would be. If we use `tar` it will be more difficult for Windows users.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -141,6 +141,7 @@ Supported Platforms
 The batch_scoring script is tested on Linux and Windows, but it should also work on OS X. Both Python 2.7 and Python 3.x are supported.
 
 Recommended Platform
--------------------
-Python 3 is recommended over Python 2 because Python 3 seems to handle encoding errors more gracefully. Python 3 also performs better.
+--------------------
+Python 3.x is recommended over Python 2.7.x. Python 2 sometimes errors decoding datasets
+that Python 3.x handles gracefully. Python 3 is also faster.
 
--- a/README.rst
+++ b/README.rst
@@ -140,4 +140,7 @@ Supported Platforms
 -------------------
 The batch_scoring script is tested on Linux and Windows, but it should also work on OS X. Both Python 2.7 and Python 3.x are supported.
 
+Recommended Platform
+-------------------
+Python 3 is recommended over Python 2 because Python 3 seems to handle encoding errors more gracefully. Python 3 also performs better.
 
--- a/DEV_README.rst
+++ b/DEV_README.rst
@@ -48,18 +48,35 @@ Cut a release candidate
 Travis bot runs automated tests and publish new version on PyPI when
 tests are passed.
 
+
+Offline Bundle
+--------------
+
+``make offlinebundle`` will create a zip file containing:
+
+  - helper_packages/ - packages like pip and virtualenv which can be used offline
+  - required_packages/ - the batch_scoring script and its dependencies
+  - OFFLINE_INSTALL_README.txt - install instructions 
+  - get-pip.py - a script that allows us to bootstrap pip in user mode
+
+This offline install method is suitable in situations where Python 2.7 or Python 3 
+are available. ``sudo`` is not required.
+
+
 PyInstaller single-file executable - experimental
 -------------------------------------------------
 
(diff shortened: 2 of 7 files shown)


README.rst
```diff
@@ -140,4 +140,8 @@ Supported Platforms
 -------------------
 The batch_scoring script is tested on Linux and Windows, but it should also work on OS X. Both Python 2.7 and Python 3.x are supported.
 
+Recommended Platform
```
mckaymatt: I also endorse Python3 as the recommended platform.

mckaymatt:
Needs review @Axik  @smillet15  @romankolpak @nimnull @sevikkk @pprett

mckaymatt:
Here is what happens to the single-file executable when you use build on newer distro than you run on.
```
 docker run --network=none --rm -it -v ~/workspace/batch-scoring:/batch-scoring quay.io/pypa/manylinux1_x86_64  bash
Status: Downloaded newer image for quay.io/pypa/manylinux1_x86_64:latest
[root@81b39a114d44 /]# useradd -m --shell=/bin/bash -u 1001 user 
[root@81b39a114d44 /]# su user 
[user@81b39a114d44 /]$ cp -r /batch-scoring/dist/datarobot_batch_scoring_1.10.0_executables ~/
[user@81b39a114d44 /]$ cd ~/datarobot_batch_scoring_1.10.0_executables
[user@81b39a114d44 datarobot_batch_scoring_1.10.0_executables]$ ./batch_scoring --help
Error loading Python lib '/tmp/_MEIMiER9Y/libpython3.5m.so.1.0': /lib64/libc.so.6: version `GLIBC_2.14' not found (required by /tmp/_MEIMiER9Y/libpython3.5m.so.1.0)
[user@81b39a114d44 datarobot_batch_scoring_1.10.0_executables]$ ./batch_scoring_sse --help
Error loading Python lib '/tmp/_MEIjUWIkN/libpython3.5m.so.1.0': /lib64/libc.so.6: version `GLIBC_2.14' not found (required by /tmp/_MEIjUWIkN/libpython3.5m.so.1.0)
```

pypa publishes a centos5 container with several versions of python  https://github.com/pypa/manylinux 
This is probably what we should build the executable in.

PR #72, opened 2017-02-07
bootstrap with pure python wheels

mckaymatt:
This relates to the last PR but it improves how we gather wheels for the bootstrap

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,9 +52,10 @@ offlinebundle:
 		. ./TEMPVENV/bin/activate; \
 		pip install -U pip setuptools; \
 		python setup.py sdist; \
-		pip download --dest=dist/offlinebundle/helper_packages --no-cache-dir  \
+		pip download --dest=dist/offlinebundle/helper_packages --no-cache-dir --only-binary :all: \
+						--implementation=py --abi=none --platform=any \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -52,9 +52,10 @@ offlinebundle:
 		. ./TEMPVENV/bin/activate; \
 		pip install -U pip setuptools; \
 		python setup.py sdist; \
-		pip download --dest=dist/offlinebundle/helper_packages --no-cache-dir  \
+		pip download --dest=dist/offlinebundle/helper_packages --no-cache-dir --only-binary :all: \
+						--implementation=py --abi=none --platform=any \
 						pip setuptools virtualenv virtualenvwrapper wheel appdirs \
-						pyparsing six packaging ; \
+						pyparsing six packaging pbr ; \
```
nimnull: can you briefly explain the purpose for including pbr for us? 
I didn't see `pbr=True,` activation flag and pbr-prefixed options

mckaymatt: The only reason I included it is because it was a dependency for a different package and pip complained when it couldn't find it.

mckaymatt: It doesn't seem like it is needed. I can remove it.

PR #73, opened 2017-02-11
Refactor offlinebundle/pyinstaller for limited release

mckaymatt:
What this PR does / State of offline install : 
- Refactors method for building a `offlinebundle`. I moved a lot of this out of the Makefile because it just didn't make sense
  -  A lot of effort was made to ensure this works cross platform.  
  - Documentation is included in the bundle
- Refactors method for building a PyInstaller single-file-executable that should work on Linux distros newer than Centos5.
  - Shown to work on Linux and OSX, but we only have a Linux build and release process.
  - Documentation is included in the bundle
- added a dockerfile and docker-compose target for an image that will allow us to build a highly-compatible PyInstaller executable. This part is **super important** and took a large 
  portion of the time. See the comments in ``Dockerfile-pyinstaller-centos5-py35-build`` and ``DEV-README`` about  Pyinstaller.  
  - This is what should allow us to start publishing builds with some confidence. 
- I detailed a build + release process in ``DEV-README``. 
- Updates documentation to be user facing. 
- Uploaded the first releases here https://github.com/datarobot/batch-scoring/releases/tag/v1.10.0 

Not Done yet:
- Windows build
- OSX PyInstaller release. It should work but I have no idea what kind of compatibility issues there might be.  Honestly this doesn't matter as much because OSX comes with python2.7 so the offlinebundle should be a solution. 
- testing 
- Release automation. It is manual but doesn't take much time or effort
- an RPM or MSI package would be nice but this needs to come first.

Issues:
- The fact that ``batch_scoring_sse`` is a separate entrypoint basically doubles the size of the PyInstaller bundle unfortunately.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -158,10 +158,10 @@ Usage Notes
 
 Supported Platforms
 -------------------
-The batch_scoring script is tested on Linux and Windows, but it should also work on OS X. Both Python 2.7 and Python 3.x are supported.
+datarobot_batch_scoring is tested on Linux and Windows and OS X. Both Python 2.7.x and Python 3.x are supported.
 
-Recommended Platform
---------------------
-Python 3.4+ is recommended over Python 2.7.x. Python 2 sometimes errors decoding datasets
+Recommended Python Version
+--------------------------
+Python 3.4 or greater is recommended, but all versions of Python 3 should work. Python 2.7.x. will work, but it sometimes errors decoding data
 that Python 3 handles gracefully. Python 3 is also faster.
 
--- a/README.rst
+++ b/README.rst
@@ -26,6 +26,26 @@ How to install particular version: ::
 
     $ pip install datarobot_batch_scoring==x.y.z
 
+Alternative Installs
+--------------------
+
+We publish two alternative install methods on our releases_ page. These are for situations where internet is restricted or Python is unavailable.
+
+:offlinebundle:
+    For performing installations in environments where Python2.7 or Python3+ is available, but there is no access to the internet.
+    Does not require administrative privileges or pip. Works on Linux, OSX or Windows.
+    
+    These files have "offlinebundle" in their name on the release page.
+
+:PyInstaller:
+    Using pyinstaller_ we build a single-file-executable that does not depend on Python. It only depends on libc and can be installed without administrative privileges.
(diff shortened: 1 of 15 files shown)


mckaymatt:
Please Review  @datarobot/predictions cc @pprett

README.rst
```diff
@@ -142,6 +162,6 @@ The batch_scoring script is tested on Linux and Windows, but it should also work
 
 Recommended Platform
 --------------------
-Python 3.x is recommended over Python 2.7.x. Python 2 sometimes errors decoding datasets
-that Python 3.x handles gracefully. Python 3 is also faster.
+Python 3.4+ is recommended over Python 2.7.x. Python 2 sometimes errors decoding datasets
```
nimnull: should we recommend 3.5 as we have services running it?

offline_install_scripts/build_offlinebundle_dockerized.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# Usage: invoked my "make offlinebundle_dockerized"
+# it is meant to run inside a docker container where it sets up the environment
+# it then calls then runs build_offlinebundle.sh as user
+apt-get update
+apt-get install -y zip unzip -q
```
nimnull: what about cleaning apt cache after all? 
`rm -rf /var/lib/apt/lists/*`

nimnull:
Looks good except of the notes I left.

README.rst
mckaymatt: I was trying to have that read "Python 3.4 or greater". I can make that clearer

offline_install_scripts/build_offlinebundle_dockerized.sh
mckaymatt: That would make sense if we were building an image, but we are only building a container. The contain is run with `--rm` so it gets wiped out at the end.

setup.py
```diff
@@ -35,8 +35,7 @@
 setup(
     name='datarobot_batch_scoring',
     version=version,
-    description=("A script to support start/resume batch scoring "
-                 "via Datarobot API."),
+    description=("A script to score CSV files via DataRobot's prediction API"),
```
mckaymatt: FYI  @madmott

PR #79, opened 2017-05-05
Support wide fields [PRED-965]

dsakagi:
By specifying a value for csv.field_size_limit, we can overcome the
arbitary limit imposed by the csv module.

- [x] I think I need another test that it works when strung all together.
- [x] I also need to make sure that you can specify this value in the config.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,11 @@
+1.11.0 (beta)
+=============
+* New parameter `field_size_limit` allows users to specify a larger maximum field
+  size than the Python `csv` module normally allows. Users can use a larger number
+  for this value if they encounter issues with very large text fields, for example.
+  Please note that using larger values for this parameter may cause issues with
+  memory consumption.
+
 1.10.2 (2017 May 9)
 ================
 * Set default timeout on server response to infinity.
--- a/tests/test_reader.py
+++ b/tests/test_reader.py
@@ -17,7 +17,7 @@ class TestCSVReaderWithWideData(object):
     @pytest.yield_fixture
     def really_big_fields_enabled(self):
         old_limit = csv.field_size_limit()
-        csv.field_size_limit(sys.maxsize)
+        csv.field_size_limit(2 ** 28)
         yield
         csv.field_size_limit(old_limit)
 
--- a/tests/test_reader.py
+++ b/tests/test_reader.py
@@ -21,23 +21,35 @@ class TestCSVReaderWithWideData(object):
         yield
         csv.field_size_limit(old_limit)
 
-    def test_slow_reader_with_really_wide_field_fails_default(self, csv_file_handle_with_wide_field):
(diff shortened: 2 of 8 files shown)


dsakagi:
@Axik do we have the equivalent of `jarvis test BLAH` for these Travis tests? It looks like one of the tests hung, trying to figure out the best way to poke it so it will try again.

dsakagi:
@datarobot/predictions Can I get a review, por favor?

CHANGES.rst
```diff
@@ -1,3 +1,11 @@
+1.11.0 (beta)
```
Axik: why "beta"?

tests/test_reader.py
```diff
@@ -0,0 +1,54 @@
+import csv
+
+import pytest
+from mock import Mock
+
+from datarobot_batch_scoring.reader import FastReader, SlowReader
+
+
+class TestCSVReaderWithWideData(object):
```
Axik: Let's keep our test structure somewhat consistent and don't use unittest-like classes and put fixtures in conftest and use plain `test_` in test files.

CHANGES.rst
dsakagi: because what if someone else has a bugfix they want to release soon?

Axik:
It'll be just 1.10.1

dsakagi:
I don't think that's right - 1.10.2 is already on PyPI. 

Also, semantically it's a new feature that doesn't break old usage, so it seems a lot to me like it should be a major revision bump.

tests/test_reader.py
dsakagi: I used to think like this too, but organizing tests into test classes makes it easier to avoid shooting yourself in the foot with fixture names in the module.

Like, if you define a test fixture `foo` then forget to include it as a dependency of the script, your IDE won't be able to tell you that you've done something wrong because there is a `foo` object accessible in the namespace.

It's just for grouping of tests - there is no unittest style test setup or anything else like that. I promise it's an enhancement, not a regression.

CHANGES.rst
Axik: sorry, I did typo on last comment on this thread (wanted to say that next will be `1.11.1`
My comment was not about version itself (`1.11.0` is completely okay) but about this `beta` addition. We usually don't have _betas_, because this package not in the beta stage. It was released already.

tests/test_reader.py
Axik: Ok.

CHANGES.rst
dsakagi: ah, I'm not referring to the package as being in beta, but about 1.11.0 being in beta (and not released) - the others have release dates because they were released.

What I was trying to do was get us so that we could put more than one fix/upgrade into the next release instead of each of us having to do a release. Do we have an automated releaser or something that would make it so we should just always release a new version for each merged PR?

Axik: I got it. No, release automated by pushing tag.

Axik: Ok

PR #82, opened 2017-05-22
Handle case when file is too big for autosample [PRED-1240]

dsakagi:
In cases where the first few rows don't fit within 512KB, the autosampler
would fail. Now these cases will fallback to a sample size of 10.

This changes also adjusts wording that made it sound like we
were going to remove the n_samples parameter. That parameter
will continue to be used.


dsakagi:
Aww yiss.

@sevikkk you were the last one to touch the auto_sampler (even though I think it was just to separate code into the `reader.py` module) - could you review this?

PR #83, opened 2017-05-29
Fix UI logging for client errors.

Axik:
When we're moved from `api/v1` to `predApi/v1.0` we did forget change
error processing. Error message moved from "status" value to "message".


requirements.txt
```diff
@@ -1,5 +1,5 @@
 six>=1.9.0
-requests>=2.7.0
```
Axik: last version of `requests` conflicting with our pinned `chardet` package version.

PR #84, opened 2017-06-06
[PRED-1090] When sniffing CSVs, don't sniff too much. It smells bad

romankolpak:
Ticket: https://datarobot.atlassian.net/browse/PRED-1090

When we sniff a bigger chunk of data, in some cases like with `ohsumed_binary_test20.csv` dataset, the `csv.Sniffer` may get confused. The heuristic that it uses is incredibly complex, so encountering weirdness is likely when throwing too much data at it. In this particular case, the `delimiter` is guessed as `' '`, which is wrong.

I looked up how `pandas` is doing this under the hood when reading csvs, and it's also sniffing the first line only. Let's be like pandas!

cc @datarobot/predictions


datarobot_batch_scoring/reader.py
```diff
@@ -379,6 +379,8 @@ def investigate_encoding_and_dialect(dataset, sep, ui, fast=False,
     else:
         opener = open
     with opener(dataset, 'rb') as dfile:
+        line_to_sniff = dfile.readline()
```
dsakagi: @romankolpak I'm not sure you'll get enough interesting data here because this is just the header. The header will let us find the delimiter, but we probably won't find the interesting stuff like quoting until later on. Does that make sense?

Axik:
tests failed cause of lint.

datarobot_batch_scoring/reader.py
romankolpak: yeah, you're right. I double checked in pandas source and they sniff the first line only for the delimiter.

I think, given how unreliable the sniffer is for edge cases like this, we need to allow users to specify the dialect options -- delimiter, qoutechar, lineterminator, etc. We can try and sniff the csv and provide a user-friendly output of what we sniffed with a message, like "We tried to automatically detect this and this is what we've got. If this is wrong, use `--delimiter` (or other options)"

For example, `pandas.read_csv` allows you to override every dialect property. They sniff the csv only to detect the delimiter, everything else is passed as an argument with a sane default.

smillet15: What about, going back to using pandas? The core reason we moved away from it, as I understand things, was the heavyweight dependencies and the challenges getting them installed. But we have a binary executable that we ship now with all deps. So maybe using pandas again makes sense?

romankolpak:
Closing this since this is not a good solution to the problem.

PR #85, opened 2017-06-10
✨ Implement most wanted --resume option feature.

Axik:


datarobot_batch_scoring/main.py
```diff
@@ -106,28 +120,28 @@ def parse_args(argv, standalone=False):
...
     conn_gr.add_argument('--n_retry', type=int,
-                         default=3,
+                         default=defaults['n_retry'],
                          help='Specifies the number of times DataRobot '
                          'will retry if a request fails. '
                          'A value of -1 specifies an infinite '
                          'number of retries. (default: %(default)r)')
-    conn_gr.add_argument('--resume', action='store_true',
-                         default=False,
+    conn_gr.add_argument('--resume', const=True, type=str2bool,
```
ikalnytskyi: Just want to notice that among Unix command line tools it's a common practice to have two different flags to distinguish *yes*/*no*. Something like:

 * ``--resume``
 * ``--no-resume``

With this approach there's no need for `str2bool` hacks.

Axik: I wanted to avoid adding a new flag.

dsakagi: If someone was using `--resume` before, will this change break their workflow?

If so, then we have to go for the `--no-resume` route to preserve functionality.

Axik: You're right.

datarobot_batch_scoring/writer.py
```diff
@@ -73,7 +73,8 @@ def create(cls, resume, n_samples, out_file, pid, lid,
             is_resume = None
             if resume:
                 is_resume = True
-            if is_resume is None:
+            # Ask if resume option wasn't specified
+            if is_resume is None and resume is None:
```
ikalnytskyi: It's a bit confusing to have `is_resume` and `resume` at the same place. I think we can simplify everything if we write something like:

```python
         file_context = ContextFile(pid, lid, n_samples, keep_cols)

         # if resume wasn't specified and there's a context to resume
         if resume is None and file_context.exists():
            resume = ui.prompt_yesno('Existing run found. Resume')

         if resume:
             ctx_class = OldRunContext
         else:
             ctx_class = NewRunContext
 
         return ctx_class(n_samples, out_file, pid, lid, keep_cols, n_retry,
                          delimiter, dataset, pred_name, ui, file_context,
                          fast_mode, encoding, skip_row_id, output_delimiter)
```

tests/test_main.py
```diff
@@ -471,3 +471,93 @@ def test_datarobot_transferable_call(monkeypatch):
             compression=False,
             field_size_limit=None,
         )
+
+
+def test_resume_key(monkeypatch):
```
ikalnytskyi: I think it might be a bit simpler and clear if you split this test into `test_resume_option` and `test_no_resume_option`.

It'd be also nice to add another test case that tests our *resume* behavior deep inside, by detecting which context exactly is created.

tests/conftest.py
```diff
@@ -61,3 +62,11 @@ def csv_file_with_wide_dataset(csv_data_with_wide_dataset):
         f.write(csv_data_with_wide_dataset.getvalue().encode('utf-8'))
     yield f.name
     os.remove(f.name)
+
+
+@pytest.yield_fixture
+def run_context_file():
+    c_file = ContextFile('pid', 'lid', 10, None)
+    open(c_file.file_name, 'a')
```
ikalnytskyi: Shouldn't it be 

```python
with open(c_file.file_name, 'a'):
    yield c_file
```

? or `c_file.clean()` ensures proper closing of the file?

Axik: clean() only removing this file.

PR #86, opened 2017-06-13
[PRED-1395] Improve logging

romankolpak:
Ticket: https://datarobot.atlassian.net/browse/PRED-1395

Super small change to improve logging:
1) log version of the package
2) use better language for retries and log retry attempt count

Also, I noticed we don't ignore files such as the script output, logs, and shelves, so I've added those to `.gitignore`

@datarobot/predictions

### DIFF
--- a/datarobot_batch_scoring/network.py
+++ b/datarobot_batch_scoring/network.py
@@ -169,7 +169,7 @@ class Network(Worker):
                 self.ui.error(msg)
                 self.send_error_to_ctx(batch, msg)
             else:
-                self.ui.warning('retrying. batch {}, attempts left: {}'
+                self.ui.warning('retrying failed batch {}, attempts left: {}'
                                 .format(batch.id, batch.rty_cnt - 1))
                 self.push_retry(batch)
 
--- a/datarobot_batch_scoring/network.py
+++ b/datarobot_batch_scoring/network.py
@@ -169,8 +169,8 @@ class Network(Worker):
                 self.ui.error(msg)
                 self.send_error_to_ctx(batch, msg)
             else:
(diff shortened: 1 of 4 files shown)


datarobot_batch_scoring/network.py
```diff
@@ -169,8 +169,8 @@ def _response_callback(self, r, batch=None, *args, **kw):
                 self.ui.error(msg)
                 self.send_error_to_ctx(batch, msg)
             else:
-                self.ui.warning('respooling batch {}'
-                                .format(batch.id))
+                self.ui.warning('retrying failed batch {}, attempts left: {}'
```
ikalnytskyi: It's more recommended way to use logging formatting style, i.e:

```
self.ui.warning('retrying failed batch %s, attempts left: %d', batch.id, batch.rty_cnt - 1)
```

The idea is to do not initiate string formatting if logging is configured to print only certain log levels. Though, in this case it won't be a performance bottleneck and I'm ok with it. Just FYI. :)

romankolpak: fun fact: batch scoring's `UI` class doesn't support this: https://github.com/datarobot/batch-scoring/blob/87148321b07c16d9b34d6818f5c89276415d75eb/datarobot_batch_scoring/utils.py#L126-L126,
it accepts just a string and no additional args. 

but you're absolutely right.

dsakagi:
Add an `Enhancements` section to the changelog and include this - then I think you're in good shape to ship.

datarobot_batch_scoring/network.py
ikalnytskyi: I thought it's logger instance, sorry. )

PR #88, opened 2017-07-03
[PRED-1377] Fix broken functionality for python 3.6 on Windows

romankolpak:
Ticket: https://datarobot.atlassian.net/browse/PRED-1377

Creating a subprocess on Windows involves pickling the `multiprocessing.Process` object and if any attributes of the process object (e.g. `target` or `args`) have problem with `pickle`, it's going to fail with an error. In our case this was `_csv.Dialect` object, which is not picklable.

This PR fixes this my wrapping the `dialect` in process's `args` in a `namedtuple` and removing `dialect` reference from the instances of `Shovel` class.

Note about python 3.6:

I tried adding 3.6 to AppVeyor and Travis, but turns out PyInstaller doesn't support Python 3.6 officially, which means for offline installs we don't support 3.6. Otherwise the package should work fine for 3.6 on both windows and unix.

сс @datarobot/predictions

### DIFF
--- a/datarobot_batch_scoring/utils.py
+++ b/datarobot_batch_scoring/utils.py
@@ -1,8 +1,10 @@
+import csv
 import getpass
 import io
 import logging
 import os
 import sys
+from collections import namedtuple
 from functools import partial
 from gzip import GzipFile
 from os import getcwd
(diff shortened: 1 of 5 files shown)


datarobot_batch_scoring/utils.py
```diff
@@ -433,6 +435,37 @@ def get_rusage():
...
+            delimiter=dialect.delimiter,
+            doublequote=dialect.doublequote,
+            escapechar=dialect.escapechar,
+            lineterminator=dialect.lineterminator,
+            quotechar=dialect.quotechar,
+            quoting=dialect.quoting,
+            skipinitialspace=dialect.skipinitialspace,
+            strict=getattr(dialect, 'strict', False))
+
+    def to_dialect(self):
```
sevikkk: it looks hackish... Why you don't just create csv.Dialect instance passing all parameters to constructor?

romankolpak: `csv.Dialect` is my favorite class in Python standard lib from now on

PR #89, opened 2017-07-11
📚 Adds a section to README for version compat

dsakagi:
Since we had a breaking change starting with DR3.0, it will be
important for this information to be readily available to users.

re: PRED-1475

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,8 +10,8 @@ Enhancements
 
 Documentation
 -------------
-A new "Version Compatibility" section has been added to the README to help surface to users any
-incompatibilities between versions of `batch_scoring` and versions of `DataRobot`.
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -8,6 +8,11 @@ Enhancements
 * New argument `no-resume` that allows you to start new batch-scoring run from scratch without being questioned about previous runs.
 
 
+Documentation
+-------------
+A new "Version Compatibility" section has been added to the README to help surface to users any
+incompatibilities between versions of `batch_scoring` and versions of `DataRobot`.
+
```
romankolpak: nit: should this be a list item, like the items in Enhancements?

dsakagi: Yep. Good catch!

PR #94, opened 2017-07-25
Refactor worker

dsakagi:
Before I went about trying to put back support for `api/v1`, I want to make sure I understand
the pieces of the `Network` process.

This PR separates the dry_run logic out to a separate class (`DryRunNetworkWorker`) that provides
the same interface as `Network` - specifically, the same `__init__` and a custom `go` method, though some
of its methods have the same names as those in `Network`.

The pieces that were then identical between the two were factored into a `BaseNetworkWorker` class
to make it easier to see exactly what was different between the dry run work and the real-deal work.

With the dry-run work out of the way, and after having to look over every line in this module more in depth, it will be easier for me to add back in support for `api/v1` on the `Network` process. It has already been pointed out that it also affects the `Writer` process. I'll get to that later.


dsakagi:
@sevikkk what do you think of this change? Like I said, I was mostly making sure that I know how this part of the code works, so it's not 100% necessary. But I kind of like separating the code into having two separate "Network" workers. Let me know what you think.

datarobot_batch_scoring/network/base_network_worker.py
```diff
@@ -0,0 +1,209 @@
...
+    the network_deque from which they are also read and sometimes retried.
+
+    Successfully returned responses are put into the writer_queue for
+    processing by another worker.
+
+    Occasional progress updates are occasionally put into the progress_queue
+    for processing by another worker
+
+    A properly functioning Worker implementation must read data from the
+    network_queue until it finds the SENTINEL message, which is the indicator
```
sevikkk: Sentinel is no longer used really as it does not work with multiple workers.

Actually worker just indicates idle state when no new requests are available in input queues, and main process monitors states of all workers and goes thru app state.

At first it waits for successful completion of Reader, after that it waits for all network workers go Idle, and waits for writer after that.

PR #95, opened 2017-08-01
Yehor/pred 1342 fix dialect detection

nimnull:
Brief intro:
Added own parser to detect features delimiter.
Made resampling to feed `csv.sniffer` only with completed lines (last line could be truncated by default)

### DIFF
--- a/datarobot_batch_scoring/detect.py
+++ b/datarobot_batch_scoring/detect.py
@@ -62,7 +62,6 @@ class Detector(object):
             if float(0.0) == deviation:
                 candidates.append(delim)
         return candidates
-            # freq struct {'': {0:0}}
 
     def mean(self, line_freq, lines_analyzed):
         """
--- a/datarobot_batch_scoring/detect.py
+++ b/datarobot_batch_scoring/detect.py
@@ -1,0 +1,81 @@
+import math
+import re
+
+
+class Detector(object):
+
+    frequency_table = None
+    non_delimiter_re = re.compile('[\w]', re.UNICODE)
+    buff_size = 1024 * 2048
+
+    def __init__(self):
+        self.frequency_table = {}
+
+    def increment(self, char, lines):
+        if char not in self.frequency_table:
+            self.frequency_table[char] = {}
+        if lines not in self.frequency_table[char]:
+            self.frequency_table[char][lines] = 0
+
+        self.frequency_table[char][lines] += 1
+
+    def detect(self, sample):
+        lines_analyzed = self.get_sample(sample)
+        return self.analyze(lines_analyzed)
+
+    def get_sample(self, sample, sample_lines=20, quotechar='"'):
+        enclosed = False
+        actual_lines = 1
+
+        # raise Exception("Sample size %d" % len(sample))
+
+        for idx, ch in enumerate(sample):
+            if actual_lines >= sample_lines:
+                break
+            prev = sample[idx-1] if idx else None
+            next = sample[idx+1] if (idx + 1) < len(sample) else None
+
+            if ch == quotechar:
+                if enclosed and next != quotechar:
+                    enclosed = False
+                elif not enclosed:
+                    enclosed = True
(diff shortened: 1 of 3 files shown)


datarobot_batch_scoring/detect.py
```diff
@@ -0,0 +1,91 @@
...
+        resampled = "".join(lines_content[:-1])
+        return self.analyze(lines_analyzed), resampled
+
+    def get_sample(self, sample, sample_lines=20, quotechar='"'):
+        enclosed = False
+        actual_lines = 1
+        lines = []
+        single_line = ""
+
+        # raise Exception("Sample size %d" % len(sample))
```
Axik: unclear what this comment line stands for

datarobot_batch_scoring/detect.py
```diff
@@ -0,0 +1,91 @@
...
+
+        for idx, ch in enumerate(sample):
+            if actual_lines >= sample_lines:
+                break
+            prev = sample[idx-1] if idx else None
+            next = sample[idx+1] if (idx + 1) < len(sample) else None
```
Axik: built-in function re-assignment 👎 ,
especially when we're using this built-in in other part of application, like reader.

datarobot_batch_scoring/detect.py
```diff
@@ -0,0 +1,91 @@
+import math
+import re
+
+
+class Detector(object):
```
Axik: small doc string about what this class supposed to do would be awesome.
It is not very clear does it detect delimiters or encoding, for example

datarobot_batch_scoring/detect.py
```diff
@@ -0,0 +1,91 @@
...
+                single_line = ""
+
+            elif not enclosed and not self.non_delimiter_re.match(ch):
+                self.increment(ch, actual_lines)
+
+        return actual_lines, lines
+
+    def analyze(self, lines_analyzed):
+        candidates = []
+        for delim, freq in list(self.frequency_table.items()):
```
Axik: I thought items() returning iterable in both cases, and we don't need to list evaluate it.

romankolpak:
I'm a bit concerned about introducing our own heuristic to detect dialects to be honest. Automatic detection of dialects is a hard problem, especially for larger datasets. 

Maybe we could provide a more fine-grained control over the dialect properties to the user instead? An example of that would be `pandas.read_csv` which provides an ability to override stuff like `qoutechar` and `lineterminator` via its arguments. I feel like that's a more robust solution, as we are unlikely to cover all the edge-cases with home-grown detector. 

Let me know what you think.

datarobot_batch_scoring/detect.py
nimnull: debug, did not dropped it yet

nimnull:
@romankolpak 
If you read this code clearly, I just introduced resampling to feed csv.Sniffer with complete set of lines. Without truncated one. Also. My detector works only as a doom's day machine, when automatic detection failed. 
Both approaches use only a limited size sample for detection. So this is no difference between 10Mb and 1T files. 
Pandas is prohibited due to implicit column type conversions. We did parse with pandas many versions earlier, but abandoned that approach because of that 'types' reason.

datarobot_batch_scoring/detect.py
nimnull: it is obvious, cause there is no calls to `chardet` and it used at the point, when we already know data encoding (:. 
Just delimiters. Also, it's main method return value called `delimiter`.

datarobot_batch_scoring/detect.py
nimnull: yep, nice thought. Missed to clean that part after refactored resampling.

datarobot_batch_scoring/detect.py
Axik: I mean, you have to thoroughly read the code, comments/docs string made for common-sense obvious information that gives heads up before code readeing.

PR #96, opened 2017-08-09
Trafaret pin

dsakagi:
With the 0.11.0 release, the use of the regex parameter in the String
type was unceremoniously removed.

We will pin to 0.10.0 now and release; we will have to deal with upgrading later. I hope he left a clear upgrade path.


dsakagi:
@Axik if you're okay with this, can you merge it and then make the new release? Looking forward to the documentation so we don't have to bother you again :)

PR #98, opened 2017-08-15
Fix manifest file directory name [PRED-1592]

dsakagi:
I don't know if this directory was renamed at some point or if this
manifest has always been wrong, but this line in the manifest seems
an awful lot like it wants to include all the source file of the
package... but doesn't.

But now it does!


dsakagi:
Tested by:
```
workon batch-scoring
python setup.py sdist
cd
mkvirtualenv testbs
pip install ~/workspace/batch-scoring/dist/datarobot_batch_scoring-1.12.1.tar.gz
python
>>> from datarobot_batch_scoring.network import Network
```

setup.py
```diff
@@ -42,7 +42,8 @@
     maintainer_email='support@datarobot.com',
     license='BSD',
     url='http://www.datarobot.com/',
-    packages=['datarobot_batch_scoring'],
+    packages=['datarobot_batch_scoring',
+              'datarobot_batch_scoring.network'],
```
ikalnytskyi: I'd suggest to use [`setuptools.find_packages()`](http://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages) in order to prevent similar issues in the future. It simply recursively walks and find packages to install.

E.g.:

```python
from setuptools import setup, find_packages


packages=find_packages(exclude=['tests*'])
```

smillet15: @ikalnytskyi Maybe file an improvement ticket for that?

PR #99, opened 2017-08-15
[PRED-1593] Remove the -e from the pip installs for the package

smillet15:
Using an editable install was hiding a bug when the package was
refactored but the new sub-library was not being properly shipped.

### DIFF
--- a/Dockerfile-3.5
+++ b/Dockerfile-3.5
@@ -8,7 +8,7 @@ RUN pip install -r requirements.txt -r requirements-test.txt
 
 COPY . /opt/project
 
-RUN pip install -e .
+RUN pip install .
 
 COPY docker-entrypoint.sh /
 
--- a/Dockerfile-3.5
+++ b/Dockerfile-3.5
@@ -8,7 +8,7 @@ RUN pip install -r requirements.txt -r requirements-test.txt
 
 COPY . /opt/project
 
-RUN pip install .
+RUN pip install -e .
 
 COPY docker-entrypoint.sh /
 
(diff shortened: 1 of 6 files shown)


smillet15:
So, we have evidence that the tests fail as they should. 

https://travis-ci.org/datarobot/batch-scoring/builds/264599687?utm_source=github_status&utm_medium=notification
https://travis-ci.org/datarobot/batch-scoring/builds/264599611?utm_source=github_status&utm_medium=notification
https://ci.appveyor.com/project/Axik/batch-scoring/build/960
https://ci.appveyor.com/project/Axik/batch-scoring/build/961

Will rebase and I expect green with the PR that was just merged to master.

Dockerfile-3.5
```diff
@@ -8,7 +8,7 @@ RUN pip install -r requirements.txt -r requirements-test.txt
 
 COPY . /opt/project
 
-RUN pip install -e .
```
sevikkk: These dockerfiles are used for in-place debug, and have volumes configured in docker-compose files, so removing -e from them will break running from docker-compose.

The same is for docker-entrypoint file.

So I don't have any objections to Makefile change and Appveyor config, but I don't think changes to docker related things are really required

smillet15: @sevikkk Makes sense. I was just casting a wide net to get something up and confirm the failure and then fix. Will clean that up shortly.

dsakagi:
So then... we don't want to do this?

smillet15:
@dsakagi We want the appveyor and Makefile changes still since they will actually fix the underlying bug and make sure the package installs correctly for future changes.

smillet15:
And on travis removing the `-e` breaks coverage? Hmmm.

smillet15:
@sevikkk Can you take a look again when you have a moment. Seems like the appveyor tests are the only place we can drop the `-e` but that gives me enough confidence in catching wholesale packaging errors in the future.

PR #100, opened 2017-08-23
[PRED-1532] set doublequote to True, if escapechar is not detected

sevikkk:
We need one or another, otherwise writer can't write values with
quotes in them.


datarobot_batch_scoring/reader.py
```diff
@@ -406,6 +406,11 @@ def sniff_dialect(sample, encoding, sep, skip_dialect, ui):
         dialect = sniffer.sniff(resampled, delimiters=delimiter)
         ui.debug('investigate_encoding_and_dialect v2 - seconds to detect '
                  'csv dialect: {}'.format(time() - t2))
+
+    if dialect.escapechar is None:
+        csv.register_dialect('dataset_dialect', dialect,
+                             doublequote=True)
+        dialect = csv.get_dialect('dataset_dialect')
```
dsakagi: Could you add a test of `sniff_dialect` that shows this behaving as you expect?

sevikkk:
the problem is - we need pretty big dataset for this - as we need dataset without quoted data in first 10 megs, and with quotes after it..

Ok, I can generate it dynamically

dsakagi:
Seva: Master Problem Solver :crown:

dsakagi:
Well I guess that's one way to do it :)

I probably would have generated it all dynamically instead of having those three test fixtures, but I'm not opposed to shipping it like this when it goes green.

datarobot_batch_scoring/reader.py
dsakagi: n.b. it has been added

PR #101, opened 2017-08-23
Pred 1587 random quit

hughdbrown:
Greg Michaelson reported that on an OSX laptop running python 3.6, batch_scoring falls over with no error message.

The fix is to run the code in that environment with more safety restrictions. The safest seems to be to use the `multiprocessing.SyncManager`.

### DIFF
--- a/datarobot_batch_scoring/batch_scoring.py
+++ b/datarobot_batch_scoring/batch_scoring.py
@@ -85,7 +85,7 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
                                            concurrent)
 
     with ExitStack() as stack:
-        if os.name is 'nt':
+        if os.name == 'nt':
             #  Windows requires an additional manager process. The locks
             #  and queues it creates are proxies for objects that exist within
             #  the manager itself. It does not perform as well so we only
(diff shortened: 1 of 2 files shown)


ikalnytskyi:
I'm not sure I'm following the error. How does exactly `multiprocessing.SyncManager` help? What was the root cause of the issue? Why it's only occurred in python3.6 on macOS?

hughdbrown:
> What was the root cause of the issue?

The root cause is still believed to be a race condition in the code,
probably in the requests.session code. And I'd have thought that placing a
lock around access to this would help. It does not.

In the alternative, I tried out using the multiprocessing.SyncManager and
we don't get any unexplained failures out of it.

Why do I see a problem only with OSX + python 3.6? I don't know.

I got to a point that my budget for fixing it was exhausted. I had a fix
that gave stability in the case that the reporting user cared about. I went
with that over WONT_FIX.

I am open to suggestions if you can see a better fix.

sevikkk:
because you basically serialized a lot of things by common interlocked manager accessor :)

Do you have any performance measurements before and after this fix? Manager makes light-weight shared memory access to flags much more heavy-weight network based one, and we check flags a lot..

hughdbrown:
> Do you have any performance measurements before and after this fix?
It is observably slower. It is a tradeoff of performance for stability.

PR #102, opened 2017-08-25
[PRED-1491] Refactor WriterProcess to make it easier to support other versions of Pred API

romankolpak:
I want to split work for https://datarobot.atlassian.net/browse/PRED-1491 in 2 separate PRs, one for refactoring, and second for actually implementing the support for v1 predictions.

This PR is the refactoring one. I tried making commit messages informative for easier review.

cc @datarobot/predictions

### DIFF
--- a/datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
+++ b/datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
@@ -1,0 +1,76 @@
+import operator
+import json
+
+
+def format_data(result, batch, **opts):
+    pred_name = opts.get('pred_name')
+    keep_cols = opts.get('keep_cols')
+    skip_row_id = opts.get('skip_row_id')
+    fast_mode = opts.get('fast_mode')
+    input_delimiter = opts.get('delimiter')
+
+    single_row = result[0]
+    fields = sorted(record['label']
+                    for record in single_row['predictionValues'])
+
+    if pred_name is not None:
+        out_fields = ['row_id', pred_name]
+        # batch scoring returns only positive class for binary tasks if
+        # user specified pred_name option, so we eliminate negative
+        # result
+        if len(fields) == 2 and single_row['prediction'] in fields:
+            fields = [fields[-1]]
+    else:
+        out_fields = ['row_id'] + fields
+
+    pred = [[record['rowId'] + batch.id] + [
+        pred_vals['value'] for pred_vals in
+        sorted(record['predictionValues'],
+               key=operator.itemgetter('label'))
+        if pred_vals['label'] in fields]
+            for record in sorted(result,
+                                 key=operator.itemgetter('rowId'))]
+
+    if keep_cols:
+        # stack columns
+
+        feature_indices = {col: i for i, col in
+                           enumerate(batch.fieldnames)}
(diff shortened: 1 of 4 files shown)


datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
```diff
@@ -0,0 +1,76 @@
...
+            written_fields = out_fields
+        else:
+            written_fields = out_fields[1:]
+            comb = [row[1:] for row in pred]
+
+    return written_fields, comb
+
+
+def unpack_data(request):
+    exec_time = float(request['headers']['X-DataRobot-Execution-Time'])
```
nimnull: what about UI debug messages? You eliminated it here but we may need that output for debugging purpose since it was declared in original flow

nimnull:
```
self._ui.debug(('successful response {}-{}: exec time {:.0f}msec | '
 -                        'round-trip: {:.0f}msec'
 -                        '').format(batch.id, batch.rows, exec_time,
 -                                   elapsed_total_seconds * 1000))
```
This message was dropped

nimnull:
LGTM if debug data would be returned to it's original place.

datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
romankolpak: it's still there, only moved outside of the scope of this function, see: https://github.com/datarobot/batch-scoring/pull/102/files#diff-b94ae56c63cbdb60b05cddbca4ff0193R415

nimnull: Got it, thanks. There is a bit complex to follow all things =(

dsakagi:
Hey @romankolpak I think you might need to adjust the `setup.py` script to include your new package (api_response_handlers). I think @ikalnytskyi recommended that we use `setuptools.find_packages` - might be a good thing to do now.

romankolpak:
Ahh, I was wondering why CI was failing with import error. Thanks, will do!

smillet15:
🎉 for tests catching this.

dsakagi:
:fireworks: :+1: :100: for tests catching this

datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
```diff
@@ -0,0 +1,76 @@
...
+    else:
+        out_fields = ['row_id'] + fields
+
+    pred = [[record['rowId'] + batch.id] + [
+        pred_vals['value'] for pred_vals in
+        sorted(record['predictionValues'],
+               key=operator.itemgetter('label'))
+        if pred_vals['label'] in fields]
+            for record in sorted(result,
+                                 key=operator.itemgetter('rowId'))]
```
hughdbrown: Still really hard to read.

romankolpak: Yeah, I know. I'd like to limit the scope of this PR to just moving code around. I'll add some unit tests to this code area in https://github.com/datarobot/batch-scoring/pull/105 so it's going to be easier to refactor this afterwards.

dsakagi:
:crossed_fingers:

PR #104, opened 2017-08-25
Roman/ci test

romankolpak:
just making sure the CI on my other pr is not playing tricks on me


romankolpak:
not playing tricks on me it seems

PR #105, opened 2017-08-28
Add support for API v1 predictions

romankolpak:
This PR is a branch off of https://github.com/datarobot/batch-scoring/pull/102 so I'll be merging it after that one goes in. 

This might be a bit hard to review before https://github.com/datarobot/batch-scoring/pull/102 is in since the diff includes changes from both PRs


README.rst
```diff
@@ -27,6 +27,7 @@ batch_scoring_version DataRobot Version
 --------------------- -----------------
 <=1.10                2.7, 2.8, 2.9
 >=1.11                3.0, 3.1+
```
dsakagi: you will probably need to change this line to say `>=1.11, <1.13`

CHANGES.rst
```diff
@@ -1,3 +1,12 @@
+1.13.0 (unreleased)
+===================
+
+Enhancements
+------------
+* Adds support for legacy predictions (api/v1) and a new parameter for specifying api version (--api_version)
```
dsakagi: Can you say "Returns support"?

Also, if you're not going to specify which options are available for `--api_version` here, we should tell them to look at the help page for more information.

README.rst
```diff
@@ -122,6 +123,7 @@ The following table describes each of the arguments:
  datarobot_key=<datarobot_key>    \-         \+     An additional datarobot_key for dedicated prediction instances. This argument is required when using on-demand workers on the Cloud platform, but not for Enterprise users.
  password=<pwd>                   \-         \+     Specifies the password used to acquire the API token. Use quotes if the password contains spaces. You must specify either the password or the API token argument. To avoid entering your password each time you run the script, use the ``api_token`` argument instead.
  api_token=<api_token>            \-         \+     Specifies the API token for requests; if you do not have a token, you must specify the password argument. You can retrieve your token from your profile on the **My Account** page.
+ api_version=<api_version>        \+         \+     Specifies the API version for requests. If omitted, defaults to current latest. Override this if your DataRobot distribution doesn't support the latest API version.
```
dsakagi: We need to specify the valid options for this

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -265,7 +264,11 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
 
         exit_code = None
 
-        response_handlers = (unpack_data, format_data)
+        try:
+            response_handlers = get_response_handlers_from_url(base_url)
+        except ValueError as e:
+            ui.fatal(str(e))
```
dsakagi: I'm only looking at the diff, so forgive me if I missed something.

It looks like no matter what you specify for `--api_version`, it's just going to try to figure it out from the base_url - which doesn't seem like what we want.

tests/test_utils.py
```diff
@@ -328,6 +329,15 @@ def test_parse_host_success():
     test_parse_host('https://127.0.0.1/', 'https://127.0.0.1/predApi/v1.0/')
 
 
+def test_get_endpoint():
+    assert get_endpoint(
+        'https://dr.com/', 'api/v1') == 'https://dr.com/api/v1/'
+    assert get_endpoint(
+        'https://dr.com', 'predApi/v1.0') == 'https://dr.com/predApi/v1.0/'
+    with pytest.raises(ValueError):
+        get_endpoint('localhost', 'api/v2.0')
```
dsakagi: can you break these up into three different test cases?

CHANGES.rst
dsakagi: We also didn't mention that we will try to infer the api_version based on their endpoint url

datarobot_batch_scoring/batch_scoring.py
romankolpak: We don't infer the api version from the user's input, even though the code might be making such impression. Here's what happens:

`batch_scoring --host=http://dr.com --api_version=api/v1` results in `base_url` being `http://dr.com/api/v1` which is then passed to the main `run_batch_predictions` function. 

I could get the `response_handlers` earlier where arguments are parsed, but that'd mean we need to pass `response_handlers` down to `run_batch_predictions`. Adding another parameter to that function is something I wanted to avoid doing, given how unwieldy it already is. Instead I made it dynamically determine the `response_handlers` by looking at the `base_url` parameter it got.

So, instead of 
```python
# dumb function needs to get everything from parameters
run_batch_predictions(base_url, response_handlers) 
```
I made it

```python
# smart function! knows how to handle different api versions in base_url parameter
run_batch_predictions(base_url) 
```

A bit of extra work in exchange for not touching interface of `run_batch_predictions`. Let me know if this doesn't sound like a good reason to you.

dsakagi:
Ahhhh.... I see. I was confusing the DR API convention (where `api/v2` is part of the host/endpoint) with the batch scoring convention (where it is not).

PR #106, opened 2017-08-29
Monkey patch

hughdbrown:
https://datarobot.atlassian.net/browse/PRED-1587 implemented a quick hack to address a failure in batch-scoring. This code addresses the problem more directly.

The code needs to be tested in python 2.7 and python 3.x because there is a use of a python 3.x library function (`functools.lru_cache`) that has no backport in python 2.7, so I provided an implementation. I will merge only after local tests show it working in both python environments with various numbers of concurrent thread counts.

### DIFF
--- a/datarobot_batch_scoring/network/network.py
+++ b/datarobot_batch_scoring/network/network.py
@@ -44,14 +44,14 @@ except ImportError:
         if no_args:
             # We were called without args
             func = setting_args[0]
-        def outer(method):
-            @functools.wraps(method)
-            def func(*args, **kwargs):
+        def outer(func):
+            @functools.wraps(func)
+            def cacher(*args, **kwargs):
                 key = tuple(args) + tuple(sorted(kwargs.items()))
                 if key not in cache:
                     cache[key] = method(*args, **kwargs)
                 return cache[key]
-            return func
+            return cacher
         return outer(func) if no_args else outer
 
 
--- a/datarobot_batch_scoring/network/network.py
+++ b/datarobot_batch_scoring/network/network.py
@@ -21,6 +21,7 @@ except ImportError:
         if no_args:
             # We were called without args
             func = setting_args[0]
+
         def outer(func):
             @functools.wraps(func)
             def cacher(*args, **kwargs):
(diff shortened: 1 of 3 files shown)


datarobot_batch_scoring/network/network.py
```diff
@@ -28,6 +28,40 @@
...
+            callable(setting_args[0])
+        )
+        if no_args:
+            # We were called without args
+            func = setting_args[0]
+        def outer(func):
+            @functools.wraps(func)
+            def cacher(*args, **kwargs):
+                key = tuple(args) + tuple(sorted(kwargs.items()))
+                if key not in cache:
```
sevikkk: There is no locking in this implementation, so it can still have threading issues :)

hughdbrown: > There is no locking in this implementation
ay ay ay. Believe me, the code is pretty crowded as is. 

Also, there is only one address resolved, to my knowledge, so it's just a read -- I don't believe you can get an error without conflicting writes.

ikalnytskyi:
I still do not fully understand what was the initial problem. @hughdbrown, could you please write a comment just near a monkey-patch with explanations why do we need it?

P.S: If the problem is that `getaddrinfo` is not thread-safe, why do we need `lru_cache`? Why just to do not monkey-patch it where we acquire some lock and call original implementation? It seems like introducing caching is not related to the bug at the first glance.

datarobot_batch_scoring/network/network.py
```diff
@@ -27,7 +28,45 @@
...
+        return outer(func) if no_args else outer
+
+
+# Monkey-patch getaddrinfo with an LRU cache to minimize conflicting calls
+from requests.packages.urllib3.util.connection import socket as r_socket
+old_getaddrinfo = r_socket.getaddrinfo
+
+@lru_cache()
+def my_getaddrinfo(*args, **kwargs):
+    return old_getaddrinfo(*args, **kwargs)
```
ikalnytskyi: I think it would be better to put `lock` here instead of putting it inside a decorator. First, there're code branches in python3 builtin `lru_cache` that are not covered with lock (maxsize is 0, or None). Second, it would be a little bit more explicit that we are really interested in acquiring lock here. What do you think?

sevikkk: We don't need lock if we are not calling getaddrinfo. So it's absolutely acceptable and preferable to put it inside of decorator

sevikkk: 3.6 have thread-safe lru_cache, so we are just matching 3.6 behaviour

PR #108, opened 2017-09-11
Infinite timeout

hughdbrown:
This gives a better solution to PRED-1629 / PRED-1587 but using an infinite timeout as the default value. It is still possible to set with any other value. An infinite timeout works better with OSX and allows more threads to operate (as a side-effect).

The changes include:
- fixes to tests
- implementation of change
- update to CHANGES.rst
- removal of monkey-patch added in previous fix


hughdbrown:
Weirdness from rebasing off master instead of pulling first.

PR #109, opened 2017-09-11
Set the default timeout to infinite

hughdbrown:
This gives a better solution to PRED-1629 / PRED-1587 but using an infinite timeout as the default value. It is still possible to set with any other value. An infinite timeout works better with OSX and allows more threads to operate (as a side-effect).

The changes include:
- fixes to tests
- implementation of change
- update to CHANGES.rst
- removal of monkey-patch added in previous fix


datarobot_batch_scoring/main.py
```diff
@@ -262,7 +262,8 @@ def parse_generic_options(parsed_args):
     resume = parsed_args.get('resume')
     compression = parsed_args['compress']
     out_file = parsed_args['out']
-    timeout = int(parsed_args['timeout'])
+    timeout = parsed_args.get('timeout')
```
ikalnytskyi: (minor): `argparse` ensures it's `int` (if passed) or `None` (if not passed). Moreover, it will always exist within `parsed_args`. So we can just write `timeout = parsed_args['timeout']` without any dances here.

ikalnytskyi:
@hughdbrown did you ensure it has no issues on macOS + python3 anymore?

hughdbrown:
@ikalnytskyi Yes, I ran it again (once!) to ensure there was no problem. No droppage with 16 threads (which is my proxy for Greg's original problem with MacOSX).

PR #110, opened 2017-09-21
PRED-1658: batch_scoring --auto_sample mini batches too small for carriage-return lined datasets

hughdbrown:
Added a few simple tests for lines terminated with \r.

### DIFF
--- a/datarobot_batch_scoring/reader.py
+++ b/datarobot_batch_scoring/reader.py
@@ -204,7 +204,10 @@ class BatchGenerator(object):
     def csv_input_file_reader(self):
         filename = self.dataset
         is_gz = filename.endswith('.gz')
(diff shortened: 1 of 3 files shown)


hughdbrown:
This last commit does a number of things:
- alters `batch_scoring` code to run large batches for python 2.7 and python 3.6 (PRED-1658)
- passes all the unit tests in both language versions

It may even work on linux and Windows.

dsakagi:
Your PR title and description don't seem to match what is actually going on here since there are changes to more than just the tests. Can you update it to lay out what you're achieving with these changes?

hughdbrown:
The task started out as adding tests to show that `batch-scoring ` could handle various line endings: \r, \n, \r\n. Then I added code to make it work correctly for batches -- with gzip/text, Windows/OSX/linux, python2.7/3.x.

PR #111, opened 2017-10-07
Add argument to skip SSL verification. [PRED-1743]

Axik:
It was requested numerous of time, so I've decided to give a try to implement it. 

cc @datarobot/predictions

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -146,7 +146,8 @@ The following table describes each of the arguments:
  auto_sample                      \+         \+     Override the ``<n_samples>`` value and instead uses chunks of roughly 2.5 MB to improve throughput. Enabled by default.
  encoding                         \+         \+     Specifies dataset encoding. If not provided, the batch_scoring or batch_scoring_sse script attempts to detect the decoding (e.g., "utf-8", "latin-1", or "iso2022_jp"). See the `Python standard encodings <https://docs.python.org/3/library/codecs.html#standard-encodings>`_ for a list of valid values.
  skip_dialect                     \+         \+     Specifies that the script skips CSV dialect detection and uses default "excel" dialect for CSV parsing. By default, the scripts do detect CSV dialect for proper batch generation on the client side.
- verify_ssl                       \+         \+     Enables SSL verification. If used with a value, e.g. ``--verify_ssl=/path/to/cert``, certificate specified by the path will be used for verification.
+ ca_bundle                        \+         \+     Specifies the path to a CA_BUNDLE file or directory with certificates of trusted Certificate Authorities (CAs) to be used for SSL verification.
+                                                    Note: if passed a path to a directory, the directory must have been processed using the c_rehash utility supplied with OpenSSL.
  no_verify_ssl                    \+         \+     Disable SSL verification.
 ============================== ========== ========= ===========
 
--- a/README.rst
+++ b/README.rst
@@ -146,6 +146,7 @@ The following table describes each of the arguments:
  auto_sample                      \+         \+     Override the ``<n_samples>`` value and instead uses chunks of roughly 2.5 MB to improve throughput. Enabled by default.
  encoding                         \+         \+     Specifies dataset encoding. If not provided, the batch_scoring or batch_scoring_sse script attempts to detect the decoding (e.g., "utf-8", "latin-1", or "iso2022_jp"). See the `Python standard encodings <https://docs.python.org/3/library/codecs.html#standard-encodings>`_ for a list of valid values.
  skip_dialect                     \+         \+     Specifies that the script skips CSV dialect detection and uses default "excel" dialect for CSV parsing. By default, the scripts do detect CSV dialect for proper batch generation on the client side.
+ skip_ssl_verification            \+         \+     Disable SSL certificates verification for HTTPS endpoints. By default SSL verification is enabled and batch_scoring throw SSLError if it's unable to verify the certificate.
 ============================== ========== ========= ===========
 
 Example::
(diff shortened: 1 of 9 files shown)


datarobot_batch_scoring/utils.py
```diff
@@ -372,10 +373,12 @@ def make_validation_call(user, api_token, n_retry, endpoint, base_headers,
...
                 r = requests.post(endpoint, headers=base_headers,
                                   data=data,
-                                  auth=(user, api_token))
+                                  auth=(user, api_token),
+                                  verify=verify_ssl)
             elif not (user and api_token):
                 r = requests.post(endpoint, headers=base_headers,
-                                  data=data)
+                                  data=data,
+                                  verify=verify_ssl)
```
dsakagi: I would rather that we expose the actual functionality of the `verify` parameter to be something where we can also allow users to point at a file that should be used to verify host identities, instead of just a boolean.

Axik: agreed. I think it's already [takes path or bool](http://docs.python-requests.org/en/master/user/advanced/?highlight=verify#ssl-cert-verification) so I'll make changes

romankolpak:
I have changed this PR to address @dsakagi's comment. Now we accept the following:
* `--verify_ssl` resulting in `verify=True`
* `--verify_ssl=/path/to/cert` resulting in `verify=/path/to/cert`
* `--no_verify_ssl` resulting in `verify=False`

Default is `True`.

Let me know how this looks @Axik @dsakagi

datarobot_batch_scoring/utils.py
```diff
@@ -57,7 +57,7 @@ def verify_objectid(value):
     OptKey('skip_row_id'): t.Bool,
     OptKey('output_delimiter'): t.String,
     OptKey('field_size_limit'): t.Int,
-    OptKey('skip_verify_ssl'): t.Bool
+    OptKey('verify_ssl'): t.Bool
```
Axik: `t.Bool | t.String` ?

romankolpak: good catch

README.rst
```diff
@@ -146,6 +146,8 @@ The following table describes each of the arguments:
  auto_sample                      \+         \+     Override the ``<n_samples>`` value and instead uses chunks of roughly 2.5 MB to improve throughput. Enabled by default.
  encoding                         \+         \+     Specifies dataset encoding. If not provided, the batch_scoring or batch_scoring_sse script attempts to detect the decoding (e.g., "utf-8", "latin-1", or "iso2022_jp"). See the `Python standard encodings <https://docs.python.org/3/library/codecs.html#standard-encodings>`_ for a list of valid values.
  skip_dialect                     \+         \+     Specifies that the script skips CSV dialect detection and uses default "excel" dialect for CSV parsing. By default, the scripts do detect CSV dialect for proper batch generation on the client side.
+ verify_ssl                       \+         \+     Enables SSL verification. If used with a value, e.g. ``--verify_ssl=/path/to/cert``, certificate specified by the path will be used for verification.
```
dsakagi: This will confuse users. The default is to enable SSL verification, so saying that its purpose is to enable SSL verification is a little weird. I think we should only support this as an option that takes an argument, the path to the cert list.

And along those lines, I think we should change its name to be something more in line with that usage, i.e. `ssl_verify_cert_path`. We should also specify that it can be a path to a single file or a directory of cert files, per http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification

romankolpak:
@dsakagi @Axik tweaked the args like this:
`--ca_bundle` to specify path to CA bundle (either a file or a directory), `--no_verify_ssl` to skip SSL verification

let me know how this looks to you.

dsakagi:
Can you add tests of the argument parser?
Also, what happens if someone specifies both of the parameters?
And one more - can you specify in the docs that the default for the `ca_bundle` is to use the system's set of trusted certificates?

romankolpak:
@dsakagi this test case https://github.com/datarobot/batch-scoring/pull/111/files#diff-3e5146a0bd380b51a3c6f1307f045616R581 answers your question (in a slightly cryptic way) about specifying both parameters

tests/test_main.py
```diff
@@ -561,5 +569,25 @@ def test_resume_no(monkeypatch):
...
+            field_size_limit=None,
+            verify_ssl=True
         )
+
+
+@pytest.mark.parametrize('ssl_argvs, expected_value', [
+    ([], True),
+    (['--no_verify_ssl'], False),
+    (['--ca_bundle', '/path/to/cert'], '/path/to/cert'),
+    (['--ca_bundle', '/path/to/cert', '--no_verify_ssl'], False)
```
dsakagi: Could you not parametrize this to make it less cryptic?

Also, this detail should be explained in the docs.

romankolpak: > Could you not parametrize this to make it less cryptic?

Do you think this so cryptic that it'd justify the copy pasting of test functions? It's true that `pytest.mark.parametrize` doesn't help readability usually, but I think I still prefer the parametrized version.

> Also, this detail should be explained in the docs.

Do users need an explanation that saying us "here's a certificate for verification, oh also don't verify anything" will result in skipping SSL verification? I think this is a pretty intuitive behavior.

dsakagi: `Using this flag will cause the argument for ca_bundle to be ignored` <- not that hard.

Also, copying and pasting of test functions is not a problem to me. Not being able to understand what situation a test is covering is a problem to me.

romankolpak:
@dsakagi does https://github.com/datarobot/batch-scoring/pull/111/commits/2e02a24f90e3a028a736b68d0afe0c2ca0404d0b help?

Axik:
LGTM. Thanks for catching it!

PR #113, opened 2017-10-26
[PRED-1785] Use t.StrBool() for config booleans

ikalnytskyi:
ConfigParser doesn't understand types, and everything it reads has str
type out of box. That means all booleans we read a failed to go through
Trafaret validation, because it expects booleans. Fortunately, Trafaret
provides so called StrBool() type that expects string as input, and
produce boolean as output. So let's use it in order to do not crash
reading boolean options from config.

PRED-1785

### DIFF
--- a/datarobot_batch_scoring/utils.py
+++ b/datarobot_batch_scoring/utils.py
@@ -54,11 +54,11 @@ config_validator = t.Dict({
     OptKey('api_token'): t.String,
     OptKey('create_api_token'): t.String,
     OptKey('pred_name'): t.String,
-    OptKey('skip_row_id'): t.Bool,
+    OptKey('skip_row_id'): t.StrBool,
     OptKey('output_delimiter'): t.String,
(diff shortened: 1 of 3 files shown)


datarobot_batch_scoring/utils.py
```diff
@@ -54,11 +54,11 @@ def verify_objectid(value):
     OptKey('api_token'): t.String,
     OptKey('create_api_token'): t.String,
     OptKey('pred_name'): t.String,
-    OptKey('skip_row_id'): t.Bool,
+    OptKey('skip_row_id'): t.StrBool,
```
romankolpak: oh, cool! I didn't trafaret has `StrBool`

ikalnytskyi: I didn't know either :) That was my first time experience with trafaret. Though initially I thought I'd need to either implement my own trafaret or perform type conversion earlier using `ConfigParser.getboolean()` method.

PR #117, opened 2018-02-07
API-3873 remove pymongo dependency as unused one

tnekdatarobot:
## Rationale
Removed pymongo dependency from test requirements as its not used

## Tests
All test were green on local machine after this change

## Jira
https://datarobot.atlassian.net/browse/API-3873

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -3,6 +3,5 @@ docutils==0.12
 pytest==2.9.0
 pytest-cov==2.2.1
 pytest-flake8==0.1
-pymongo==2.8
 Flask==0.10.1
 mock>=1.3.0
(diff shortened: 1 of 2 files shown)


sevikkk:
pymongo was transient dependency for trafaret to provide ObjectID check, probably it was pinned for some reason.

There was change to replace ObjectID with regexp, but I don't remember exact outcome from it.

requirements-test.txt
```diff
@@ -3,6 +3,6 @@ docutils==0.12
 pytest==2.9.0
 pytest-cov==2.2.1
 pytest-flake8==0.1
-pymongo==2.8
+pymongo==3.5.1
```
ikalnytskyi: Why new pin? If we need pymongo only because we use `ObjectId` trafaret, maybe better add `bson` as dependency instead of `pymongo`?

tnekdatarobot: I was trying  to check if it change travis test that is red,  but it's not related will remove it again and figure out problem with tests

PR #118, opened 2018-02-22
[API-4184] Do not install "futures" in case of Python 3

ikalnytskyi:
Since futures 3.2 [1], the package enforces to be installed only in Python 2
environments because it's basically a backport of Python 3's built-in
package. So in order to support both Python 2 and Python 3 environments, we
have to skip installation of futures package in case of Python 3.

It might look natural to use environment markers [2] to achieve this goal but
they are new and were introduced in setuptools in mid of 2017. FWIW,
setuptools on both Ubuntu Trusty and Ubuntu Xenial do not support them and
batch scoring script may be installed in pretty outdated envs. So let's do it
old-fashioned way by adding condition in setup.py.

[1] https://github.com/agronholm/pythonfutures/commit/d0393ad626d25622927bb0ed47d35ddb2f6cd321
[2] https://www.python.org/dev/peps/pep-0508/#environment-markers

Jira: API-4184

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,10 @@
+1.13.1 (unreleased)
+===================
+
+Bugfixes
+--------
+* Fix package installation in Python 3 environments.
+
 1.13.0 (2017 November 10)
 =========================
 
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+import sys
 import codecs
 import os.path
 import re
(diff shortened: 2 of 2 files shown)


ikalnytskyi:
@romankolpak I'm afraid we need to release this bugfix, as currently you are unable to install the project in Py3 env with latest setuptools. Is it hard to do? Should I also update a changelog here?

romankolpak:
@ikalnytskyi you can find the docs on how to release DEV_README: https://github.com/datarobot/batch-scoring/blob/master/DEV_README.rst#release. It's not hard at all, if everything in the guide goes well, that is ;)

If you're going to push a new release after merging this in, definitely add a CHANGELOG entry.

ikalnytskyi:
@romankolpak jeez.. so we update changelog later? when we bump a version?

romankolpak:
What I did previously was this: 

1) add a new entry to the unreleased changelog in the PR where I was making changes, e.g.:

```
1.14.0 (unreleased)
-------------------
* foo bar
```

2) make a separate PR where I replace `unreleased` with a the release date in CHANGES.rst, bump package version, commit, merge PR
3) follow the steps for releasing the package in DEV_README

PR #120, opened 2018-02-26
Build different wheels for Python 2 and Python 3

ikalnytskyi:
Since recently we have different set of dependencies for Python 2 and
Python 3 environments, hence we can't have one wheel because it may have
only one set of dependencies and no way of conditional dependencies.

So this commit removes "universal" wheels, which effectively means we
will have two wheels: one for python2, and one for python3.


CHANGES.rst
```diff
@@ -1,3 +1,10 @@
+1.13.2 (2018 February 26)
```
romankolpak: are you planning to push a new release right after this is merged?

ikalnytskyi: Yep

ikalnytskyi: Because currently we have only one wheel on PyPi (py2-py3 wheel), that's built by python3 and misses `futures` dependencies. Which means, it's impossible to install it on py2. :(

PR #124, opened 2018-06-01
Adding prediction explanations support to batch scoring.

doleks:
This pull request is ready for review.

It adds support for the prediction explanations.


nott:
I think @Axik and @sevikkk are two main batch scoring experts in the company.

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -121,10 +122,17 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
         base_headers['content-type'] = 'text/csv; charset=utf8'
         if compression:
             base_headers['Content-Encoding'] = 'gzip'
+
         if import_id:
-            endpoint = base_url + '/'.join((import_id, 'predict'))
+            endpoint = base_url + '/import_id'
```
sevikkk: this one does not look right, you have `import_id` literal instead of value

datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
```diff
@@ -66,6 +67,16 @@ def format_data(result, batch, **opts):
             written_fields = out_fields[1:]
             comb = [row[1:] for row in pred]
 
+    if 'reasonCodes' in single_row:
+        num_reason_codes = len(single_row['reasonCodes'])
```
sevikkk: Do we always have the same number of reasonCodes for all rows?

AFAIK we don't generate them for all rows and number of important features can be different for different rows..

And we don't have thresholds in request arguments - does it mean we calculate them for all rows or use default project settings?

sevikkk:
Tests failed because of flake8 style checks, not because of actual test failures. It have not even run them...

datarobot_batch_scoring/batch_scoring.py
doleks: You are right. Copy/pasta failure.

datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
doleks: When the reason codes are requested, we generate reason codes for all fields, sort them, and then return only the requested number of them.

datarobot_batch_scoring/reader.py
```diff
@@ -584,6 +586,11 @@ def auto_sampler(dataset, encoding, ui):
              'recommended lines per sample: {}'.format(csv_lines, avg_line,
                                                        lines_per_sample))
     ui.info('auto_sampler: total time seconds - {}'.format(time() - t0))
+    if max_prediction_explanations:
+        # The prediction explanation scores 10 extra rows for each column for each row.
+        # But not more than for 30 columns. So the slowdown cannot be more than 300x.
```
Axik: slowdown of what? I don't get this comment.

sevikkk: @doleks this code is not supposed to limit number of lines per batch based on runtime, it just limits batch size to avoid frontend level reject of payload

2 MB payload on some hardcore blender can hit server-side timeout easily, but there is no code in batch-scoring to avoid this. If you want to implement something like this, we need to start with small batches and scale based on actual prediction time, passing suggested batch size via shared memory variable from `network` or `writer` to `reader`

doleks: Bollocks! I thought it is some time-based heuristic.

Instead of implementing the retry mechanism, it probably would be better to use scoring benchmarks that the core engine does. But in any case, it is too elaborate for what I'm trying to do. So I will remove my change.

sevikkk: I didn't mean to retry failed requests, but instead start with small batches and gradually increase batch size to meet some predefined prediction time like 10 seconds per batch

doleks: You are right. But measuring time is not deterministic either. It is prone to fluctuations based on the load. And because your proposal is self-adjusting, you need to think about all possible scenarios. Like many scripts being fired at once.
Not today.

datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
```diff
@@ -66,6 +66,22 @@ def format_data(result, batch, **opts):
...
+        for num in range(1, num_reason_codes + 1):
+            written_fields += [
+                'explanation_{0}_feature'.format(num),
+                'explanation_{0}_strength'.format(num)
+            ]
+        for in_row, out_row in zip(sorted_rows, comb):
```
pprett: if sorted_rows can be large i'd prefer the izip here but that's just a nitpick

doleks: Sure.

doleks: @pprett It seems that there is no easy way to do it in Python2/3 compatible mode. So I'll leave it as-is.

datarobot_batch_scoring/batch_scoring.py
```diff
@@ -121,10 +122,18 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
...
             base_headers['Content-Encoding'] = 'gzip'
+
         if import_id:
-            endpoint = base_url + '/'.join((import_id, 'predict'))
+            endpoint = base_url + import_id
+        else:
+            endpoint = base_url + '/'.join((pid, lid))
+
+        if max_prediction_explanations:
+            endpoint += '/reasonCodesPredictions?maxCodes=' + \
```
madmott: this is baking in using the deprecated "reasonCodesPredictions" routes instead of the "predictionExplanations" routes.  Should we add a check that tried to see if predictionExplanations exists and uses that?

datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
Axik: @doleks 

`from six.moves import zip`
https://pythonhosted.org/six/#module-six.moves

datarobot_batch_scoring/batch_scoring.py
doleks: Will change it.

datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
doleks: @Axik thank you for it.

ahazat:
+1

iahazat:
+1

pprett:
Please add a fail fast for the old (`--api_version "api/v1"`) routes that dont support prediction explanations. Currently it will give this:
```
(scb3) peter@peter-ThinkPad-P50 ~ $ batch_scoring --api_version="api/v1" ...
MainProcess [INFO] version: 1.14.2
MainProcess [INFO] platform: linux 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609]
MainProcess [INFO] Will be using API endpoint: https://cfds.orm.datarobot.com/api/v1/
MainProcess [ERROR] Authorization failed -- Please check project id and model id permissions: <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>404 Not Found</title>
<h1>Not Found</h1>
<p>The requested URL was not found on the server.  If you entered the URL manually please check your spelling and try again.</p>
.
If you need assistance please send the log file/s:
	/home/peter/datarobot_batch_scoring_main.log
to support@datarobot.com.
```

pprett:
Seems travis on python 2.7 stalled -- all tests are green -- i'll kick it again.

pprett:
jenkins please test continuous-integration/travis-ci/pr

PR #125, opened 2018-06-12
batch_scoring double prediction rows when double carriage return at line ending

devalentino:
If an input csv file contains extra end-line characters (e.g. \r\r\n), the batch-scoring script will produce extra prediction rows.  This fix resolves that issue by ignoring empty rows in the input csv


devalentino:
@romankolpak please review this

romankolpak:
I think this is a somewhat a non-standard behavior to silently skip the empty lines. Other CSV-reading software (such as Numbers on mac) doesn't do this and will render empty rows in the table, which I think is a more logical outcome for such input from user.

If we're confident this is how batch scoring should behave, I suggest adding a warning in the script output to let users know about this, e.g. `Empty lines were detected in the input file and were ignored`.

romankolpak:
Turns out that the behavior of discarding empty rows is consistent with how DataRobot reads CSV, so this should be fine. Let's add a warning to user for better UX and ship it!

PR #126, opened 2018-06-15
Include README.rst into distribution as long description

nott:
Otherwise the project looks bad at [PyPI](https://pypi.org/project/datarobot_batch_scoring/).


nott:
@tnekdatarobot I'll do that in a subsequent PR

Thank you for the review!

PR #132, opened 2018-08-01
[API-5339] Add deployment aware routes

jettify:
# BATCH SCORING PULL REQUEST


datarobot_batch_scoring/batch_scoring.py
```diff
@@ -74,7 +74,8 @@ def run_batch_predictions(base_url, base_headers, user, pwd,
                           output_delimiter=None,
                           max_batch_size=None, compression=None,
                           field_size_limit=None,
-                          verify_ssl=True):
+                          verify_ssl=True,
+                          deployment_id=None):
```
tnekdatarobot: nit: can we do pid, lid as optional?

jettify: I can try, but pid, lid are positional arguments they may break all tests

tnekdatarobot: Ok let leave it as it is

jettify:
Added one more test case and updated README.

jettify:
@tnekdatarobot is it good to go?

jettify:
Updated readme, also fixed config file handling for the new command.

jettify:
@datarobot/batch-scoring please review

datarobot_batch_scoring/batch_scoring.py
samuelclark: so would they have to pass in a pid / lid and have them be ignored, this seems confusing!

datarobot_batch_scoring/utils.py
```diff
@@ -39,6 +39,7 @@ def verify_objectid(value):
     OptKey('host'): t.String,
     OptKey('project_id'): t.String(regex='^[A-Fa-f0-9]{24}$'),
     OptKey('model_id'): t.String(regex='^[A-Fa-f0-9]{24}$'),
+    OptKey('deployment_id'): t.String(regex='^[A-Fa-f0-9]{24}$'),
```
samuelclark: why dont we just use the builtin trafaret for objectids :)

jettify: No clue, probably question for @datarobot/batch-scoring

pprett: We want to avoid dependencies as much as possible; `bson` or `pymongo` are too heavy. I would prefer the way you did it above.

PR #134, opened 2018-08-09
Use dumb dbm driver for all platforms

nimnull:
## RATIONALE

https://datarobot.atlassian.net/browse/API-4967

`shelve` uses `dbm` implemented by a number of platform-dependent C extensions. Issue reproduced several times on MacOS with `dbm.ndbm` extension. However linux with dummy (pure-python fallback) that works for all cases.

## CHANGES

Manually specify `dbm` implementation type to make it platform-issues free.

## TESTING

Ticket contains sample continuously failing on MacOS at 6599 batches (1 row per batch) even if sample is shuffled. That change allows to score that full sample with no changes to the options.


datarobot_batch_scoring/writer.py
```diff
@@ -10,6 +10,11 @@
 from functools import reduce
 from time import time
 
+try:
```
Axik:
```python
if six.PY3:
    from dbm import dumb as dumb_dbm
else:
    import dumbdbm as dumb_dbm
```

nimnull:
@Axik how about aprove?

Axik:
I missed couple of things.
1. Bump version
2. Update CHANGES.rst

nimnull:
@Axik done. Ready to merge & release

PR #136, opened 2018-08-14
[API-5329] Yehor/api 5329 added bs proxy support

nimnull:
## RATIONALE

Brought back env proxy settings support since we use overrided flow present in requests.request method.

Also uncovered and fixed issue when BS raises an exception on 0 requests made


nimnull:
@Axik case added

datarobot_batch_scoring/network/network.py
```diff
@@ -106,7 +109,8 @@ def _request(self, request):
 
         try:
             prepared = self.session.prepare_request(request)
-            self.session.send(prepared, timeout=self._timeout)
+            proxies = get_environ_proxies(prepared.url)
```
doleks: Can you add a comment to the documentation explaining that we are trying to use the right proxy settings? Also tell that it still does not work on Windows, if the proxy is set up using Auto-config https://en.wikipedia.org/wiki/Proxy_auto-config`?

doleks: Despite being truly ugly technology, the auto-config is wildly popular in the enterprises. I keep encountering it with the DataRobot customers.

nimnull: @doleks I agree with it nut let's create a separate ticket to implement windows autoconfig support.

doleks: For now, I suggest that we just document in the user-visible .readme that batch scoring will try to use proxy and document the clear case, when it does not work.

PR #138, opened 2018-08-28
 Add batch_scoring_deployment_aware.py wrapper for offline install on Windows/Linux/MacOSX

nott:
The pull request that added deployment awareness (#132) didn't contain a wrapper script for Windows/Linux/MacOSX offline install. Let's fix it and release a new minor version.

Fixes #1


Axik:
I doubt that we need to bump the version of the package since it doesn't change anything in the behaviour, but it's clear release logic fix. 

Also, I'm not sure if you want to do it in the another PR or this one, but Linux assets also misses deployment_aware executables.

nott:
> I doubt that we need to bump the version of the package since it doesn't change anything in the behaviour, but it's clear release logic fix.

I prefer not to overwrite already released packages - it's hard to tell which version is installed in that case.

nott:
> Also, I'm not sure if you want to do it in the another PR or this one, but Linux assets also misses deployment_aware executables.

I think this PR contains all the required changes. I should have mentioned this in the description. I'll double check after releasing.

PR #139, opened 2018-10-04
[API-5328] Adding check to detect quoted delimiters during --fast mode with --keep_cols

daveblumstein:
# BATCH SCORING PULL REQUEST
 
## RATIONALE

The parsing under --fast mode with --keep_cols does not support the case of quoted delimiters (like comma and newline) within table cells. This change throws an explicit exception when that seems to be happening. 

## CHANGES

* moved exceptions into own file
* added special exception and error message for the case where the parsing during --fast mode and --keep_cols results in an unexpected number of columns
* added test cases to exercise the new logic


nott:
@daveblumstein could you please update CHANGES.rst?

Other than that - LGTM

nott:
@pprett does this look good to you?

pprett:
Very nice indeed -- looks so clean

PR #140, opened 2018-11-09
Ensure new enough requests to resolve CVE-2018-18074

smillet15:
https://nvd.nist.gov/vuln/detail/CVE-2018-18074

# BATCH SCORING PULL REQUEST


pprett:
Test errors seem strange - i will do a calibration run of master

pprett:
Seems like the new chardet detects the encoding for `tests/fixtures/windows_encoded.csv` as `windows-1252` instead of the assumed `ISO-8859-2` -- strangly, if i run chardet on the cmd it will give me the correct encoding (iso). 
I checked the input to the chardet.detect method and the hash matches master so the difference is in chardet's universal detector. 

Accoriding to wikipedia:
> However a few of them are rearranged (unlike Windows-1252, which keeps all printable characters from ISO-8859-1 in the same place).

pprett:
@Axik the python 3.4 error on macpython is due to an old version of setuptools it seems

smillet15:
Thanks @Axik and @pprett!

PR #146, opened 2018-12-17
[PRED-2150] fix verify_ssl argument in api token retrieve

ajbeairsto:
# BATCH SCORING PULL REQUEST

## RATIONAL
We fail to use the verify argument in the request to get the user's api token. This means that if SSL is enabled, the request to get the API token will fail. This should fix that.


pprett:
LGTM

please add your change to https://github.com/datarobot/batch-scoring/blob/master/CHANGES.rst and then lets release this!

pprett:
See https://github.com/datarobot/batch-scoring/blob/master/DEV_README.rst#release for the release process

ajbeairsto:
Thanks @pprett!

PR #153, opened 2019-03-19
PRED-2176 refactor format_data to use generators approach

rrader:
# BATCH SCORING PULL REQUEST

## RATIONALE

The structure of `datarobot_batch_scoring.api_response_handlers.pred_api_v10.format_data` function makes it difficult to understand the code and add new columns.

## CHANGES

 - `format_data` was refactored in favor of generators approach.
 - now all columns are independent functions which return headers and generator of rows

## TESTING

 - unit tests should confirm refactoring didn't break the contract


datarobot_batch_scoring/api_response_handlers/pred_api_v10.py
```diff
@@ -1,76 +1,207 @@
...
+    prediction_explanations_key : str
+        key of results by which explanation fields can be found
+
+    Returns
+    -------
+    header: list[str]
+    row_generator: iterator
+    """
+    headers = []
+    single_row = result_sorted[0]
```
chrj: Will there always be at least one item in `result_sorted`? Otherwise this could raise an `IndexError`.

rrader: Yes, there's a check that forbids sending empty dataset for scoring much earlier
https://github.com/datarobot/batch-scoring/blob/master/datarobot_batch_scoring/reader.py#L255

PR #155, opened 2019-03-21
[PRED-2176] Add optional prediction threshold field to out csv

rrader:
# BATCH SCORING PULL REQUEST

## RATIONALE

After #154 we have new prediction decision field in the output CSV, but it's not obvious how it is calculated from probabilities of each class. Threshold, which is 0.5 by default, may be changed in deployment, so the new field to expose the threshold to output CSV was added.

## CHANGES

 - New command line argument `--pred_threshold=<column name>` which is disabled by default
- If the argument pred_threshold is given, a new column in output CSV appears with the `threshold` value. May be applied only for binary classification, otherwise an error will be showed.

## TESTING

 - Functional tests were written
 - I ran script with and without new argument against real DR cluster for binary classification.


datarobot_batch_scoring/main.py
```diff
@@ -213,6 +213,11 @@ def parse_args(argv, standalone=False, deployment_aware=False):
                         'empty name is used if not specified. For binary '
                         'predictions assumes last class in lexical order '
                         'as positive')
+    csv_gr.add_argument('--pred_threshold', type=str,
+                        nargs='?', default=None,
+                        help='Specifies column name for prediction threshold'
```
andrii-rubtsov: Should have trailng space :)

andrii-rubtsov: And, as agreed, please fix the docs below to smth like "column will be present in output, if not specified"

PR #161, opened 2019-07-05
[PRED-2644] Fix decoding error in case of dialect detection

falkerson:
# BATCH SCORING PULL REQUEST

## RATIONALE
We open some file in binary mode and read some N bytes to detect encoding. Later we use this bytes to detect dialect but before it we decode() them into string(unicode). Because we have const number of N, it's possible that during bytes reading last character may be torn apart and then during decode() we can't identify that  character.

## CHANGES
As solution we will read first N characters instead of bytes to detect dialect. Because it's called only once no huge performance degradation expected.


datarobot_batch_scoring/reader.py
```diff
@@ -432,6 +432,19 @@ def sniff_dialect(sample, encoding, sep, skip_dialect, ui):
     return dialect
 
 
+def get_opener_and_mode(is_gz, text=False):
```
ikalnytskyi: I'm not sure it's going to work for gzipped japanese dataset. I think we need some sort of combination for is_gz and text.

falkerson: in PY3 gzip.open has encoding parameter. For PY2 please take a look here: https://github.com/datarobot/batch-scoring/pull/161/files#diff-6c240e73b54162e5fce6a481761017a2R527

datarobot_batch_scoring/reader.py
```diff
@@ -432,6 +434,25 @@ def sniff_dialect(sample, encoding, sep, skip_dialect, ui):
...
+    if is_gz:
+        if six.PY2:
+            mode = 'r' if text else 'rb'
+            return (gzip_with_encoding, mode)
+        else:
+            mode = 'rt' if text else 'rb'
+            return (gzip.open, mode)
+    else:
```
ikalnytskyi: It should work, however, I think the whole `else` can be written using `io`:

```python
mode = 'rt' if text else 'rb'
return (io.open, mode)
```

because:

* `io.open` uses universal newlines by default (no need to pass `U` in openning mode)
* `io.open` is `open` in Python 3

tests/test_functional.py
```diff
@@ -125,6 +125,51 @@ def test_simple(live_server, tmpdir, func_params):
...
+        dry_run=False,
+        encoding='',
+        skip_dialect=False
+    )
+
+    assert ret is None
+    actual = out.read_text('utf-8')
+    with open('tests/fixtures/jpReview_books_reg_out.csv', 'rU') as f:
+        expected = f.read()
+    assert str(actual) == str(expected), expected
```
ikalnytskyi: How does `str` works on Python 2? I mean `.read_text` should return `unicode`, and casting a unicode to `str` should probably fail o_O I'd expect to see here `six.text_type` instead of `str`.

falkerson: doesn't matter in case of output. Here we have:
row_id,0.0,1.0
0,1.0,0.0
1,1.0,0.0
2,1.0,0.0

so it can be converted as ascii

PR #164, opened 2019-08-01
PRED-2832: Batch scoring with prediction explanations produces rows with varying column lengths

andrii-rubtsov:
Fixes the issue for prediction explanations when response rows have different number of columns depending on the number of prediction explanations in the Prediction Servers output.

This PR sets the number of rows in the output response to always be equal to `max_prediction_explanations` from the input, padding missing explanations with empty values if any.


tests/fixtures/10kDiabetes_6explanations.csv
```diff
@@ -0,0 +1,11 @@
+row_id,0,1,explanation_1_feature,explanation_1_strength,explanation_2_feature,explanation_2_strength,explanation_3_feature,explanation_3_strength,explanation_4_feature,explanation_4_strength,explanation_5_feature,explanation_5_strength,explanation_6_feature,explanation_6_strength
```
ikalnytskyi: Why `explanation_6_feature` is empty and has no value?

ikalnytskyi: Is it because we cannot compute more than 5 reasons for this project?

andrii-rubtsov: It is because the fixture contains only 5 explanations with max = 6

ikalnytskyi: I'm sorry, I didn't get it. What fixture?

andrii-rubtsov: I am going to update the fixture to have mixed number of exlanations since I need to bump version to release it anyways.

PR #171, opened 2020-01-20
[PRED-3618] Add deprecation warning re legacy Prediction API routes.

farafarafara:
# BATCH SCORING PULL REQUEST

## RATIONALE

The following Prediction API routes are deprecated:

- `/api/v1/`
- `/predApi/v1.0/<projectId>/<modelId>/`

In the future, ``/predApi/v1.0/deployments/<deploymentId>/`` will be the only supported endpoint for prediction servers.

## CHANGES

This PR adds deprecation warning when batch_scoring command is run and to the README file.


README.rst
```diff
@@ -78,21 +78,34 @@ Features
...
+Deprecation warning
+-------------------
+
+The following Prediction API routes are deprecated:
+
+- ``/api/v1/``
+- ``/predApi/v1.0/<projectId>/<modelId>/``
+
+In the future, ``/predApi/v1.0/deployments/<deploymentId>/`` will be the only supported endpoint for
+prediction servers. Contact your CFDS or Support if you need help migrating to the new API routes.
```
doleks: For end-users of this tool, it is not obvious that deprecating these API end-points will mean that we are deprecating the `batch_scoring` tool. 
I suggest rephrasing this to say something like, "The `batch_scoring` command is deprecated because the API routes (`/api/v1/` and `/predApi/v1.0/<projectId>/<modelId>/`) that are used by this command are deprecated. Instead, `batch_scoring_deployment_aware` should be used everywhere `batch_scoring` was used before."

farafarafara: Agree, rephrased.

PR #174, opened 2020-02-27
[PRED-2817] Fix failing Python3.5 tests and a bit of cleanup

dotkas:
# BATCH SCORING PULL REQUEST

## RATIONALE

There was a bug where docker-compose would fail on Python 3.5 tests, leaving developers with a bad night's sleep.

## CHANGES

The failing tests seemed to be caused by overuse of statefulness in the docker-compose files. That's bad practice. However, as we are shifting focus to the Python API Client + Batch Scoring, I will not spend too much time on this.


Makefile
```diff
@@ -76,5 +72,7 @@ clean:
 	@rm -rf datarobot_batch_scoring.egg-info build/* dist/* 
 	@rm -rf htmlcov
 	@rm -rf .coverage
+	@rm -rf .cache
 	@rm -f batch_scoring.spec batch_scoring_sse.spec batch_scoring_deployment_aware.spec
+	@rm -f *.shelve.*
```
andrii-rubtsov: Just curious what is in there?

docker-compose.yml
```diff
@@ -7,35 +7,13 @@ services:
...
-        dockerfile: Dockerfile-3.3
-     volumes:
-        - .:/opt/project
-  python34:
-     image: batch-scoring-3.4
-     build:
-        context: .
-        dockerfile: Dockerfile-3.4
-     volumes:
-        - .:/opt/project
```
andrii-rubtsov: In our README we state that we support all python 3+ versions. I think it is time to update it to 3.5+.
Maybe as a separate PR

Makefile
```diff
@@ -18,7 +18,7 @@ checkrst:
 	# page properly.
 	python setup.py check --restructuredtext
 
-test: .install .install-test-deps flake8
+test: clean .install .install-test-deps flake8
```
andrii-rubtsov: I assume this is the main contribution to fixing the problem, right?

dotkas: Yes exactly.

Makefile
dotkas: As far as I understand it is sort of like Pickle files being made by the module during runtime - it was already being git-ignored but not removed during cleaning, if I understand it correctly. https://docs.python.org/3/library/shelve.html

docker-compose.yml
dotkas: I agree 👍

